PR #963, opened 2017-10-20
Support linebreaks in require/import statements

### LINKED ISSUE #962: Newline when importing multiple members breaks deploy
When importing multiple items in a function, this works:

```
const { A, B, C } = require('../myFile');
```

This does not work:

```
const {
  A,
  B,
  C
} = require('../myFile');
```

Error:
```
graphcool deploy
 ▸    Cannot read property 'startsWith' of undefined
```

kbrandwijk:
Fixes #962.


timsuchanek:
Thanks for the PR!

PR #1040, opened 2017-10-26
Improve permission explanations

kbrandwijk:
Especially the fact that specifying fields on a permission means that the permission applies *only* to those fields.


ttlsx:
Everyone can **read** only the fields `email` and `posts` on

marktani:
thanks 🎉

PR #1057, opened 2017-10-26
fix links to examples

rosswaycaster:


rosswaycaster:
Oops, just saw a few other people had fixed this same issue.

PR #1079, opened 2017-10-29
using topPosts in final query to prevent error

josephjorgensen:
given the latest sample code, the final query in this quickstart guide was generating the following error:
`Unknown argument "limit" on field "posts" of type "User"`

this could be fixed by removing the limit, or by requesting topPosts one level up on the viewer model. i went with the latter option as it also demonstrates the limit, rather than simply returning all posts. also corrected a very minor typo in a sentence above this.


marktani:
Hey @josephjorgensen, please sign this GCA so I can merge your PR: https://cla-assistant.io/graphcool/framework?pullRequest=1079 🙂

marktani:
thanks 🎉

PR #1098, opened 2017-11-01
Local docker and DI binding cleanup

dpetrick:
- [x] Supress cloudwatch stacktraces.
- [x] Less verbose graphcool-dev output to simplify local docker logs.
- [x] Only init one connection per DB (binding errors).
- [x] Init DBs connections concurrently (temporary solution until a real approach is there).
- [x] Replace all AWS services with the corresponding mocks or remove them entirely.
- [x] Move most AWS dependencies out of the core backend and consolidate initializers.
- [x] Rework ServerExecutor: Change .startBlocking semantic to actually block the calling thread, which is required to remove AhcStandaloneClient dependency, as it created non-daemon threads that prevented the server from exiting before. Previous semantic was to init in a blocking fashion, which is confusing.
- [x] Add rudimentary HTTP client lib based on akka HTTP.
- [x] Replace (most) current akka HTTP implementations with the simple HTTP client
- [x] Remove AhcStandaloneClient dependency

Additional changes:
- Expanded .gitignore for our scala setup (adapted from our old setup).


server/single-server/src/main/resources/application.conf
```diff
@@ -34,7 +34,7 @@ logs {
   dataSourceClass = "slick.jdbc.DriverDataSource"
   connectionInitSql="set names utf8mb4"
   properties {
-    url = "jdbc:mysql:aurora://"${?SQL_LOGS_HOST}":"${?SQL_LOGS_PORT}"/"${?SQL_LOGS_DATABASE}"?autoReconnect=true&useSSL=false&serverTimeZone=UTC&socketTimeout=60000&useUnicode=true"
+    url = "jdbc:mysql://"${?SQL_LOGS_HOST}":"${?SQL_LOGS_PORT}"/"${?SQL_LOGS_DATABASE}"?autoReconnect=true&useSSL=false&serverTimeZone=UTC&socketTimeout=60000&useUnicode=true"
```
mavilein: Do we know what influence the `:aurora` part had?

server/single-server/src/main/resources/logback.xml
```diff
@@ -1,12 +1,9 @@
...
-    <!-- reference: https://github.com/slick/slick/blob/master/common-test-resources/logback.xml -->
     <logger name="slick" level="OFF" />
     <logger name="com.zaxxer.hikari" level="OFF" />
-    <!--<logger name="slick.jdbc.JdbcBackend.statement" level="DEBUG" />-->
-    <!--<logger name="slick.jdbc.StatementInvoker.result" level="DEBUG" />-->
-    <!--<logger name="slick.jdbc.JdbcBackend.benchmark" level="DEBUG" />-->
+    <logger name="cool.graph" level="INFO" />
 
-    <logger name="cool.graph" level="DEBUG" />
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
```
mavilein: what does that do?

server/single-server/src/main/scala/cool/graph/singleserver/SingleServerDependencies.scala
```diff
@@ -27,44 +26,51 @@ import cool.graph.subscriptions.protocol.SubscriptionRequest
...
+
+  val (globalDatabaseManager, internalDb, logsDb) = {
+    val internal = Initializers.setupAndGetInternalDatabase()
+    val logs     = Initializers.setupAndGetLogsDatabase()
+    val client   = Future { GlobalDatabaseManager.initializeForSingleRegion(config) }
+    val dbs      = Future.sequence(Seq(client, internal, logs))
+
+    try {
+      val res = Await.result(dbs, 15.seconds)
+      (res(0).asInstanceOf[GlobalDatabaseManager], res(1).asInstanceOf[MySQLProfile.backend.Database], res(2).asInstanceOf[MySQLProfile.backend.Database])
```
mavilein: To avoid the cast one would probably rather use a for comprehension:
```
for {
  internal <- internalFut
  logs       <- logsFut
  client     <- clientFut
} yield ...
```

server/single-server/src/main/resources/logback.xml
dpetrick: That supresses the logback status messages themselves, e.g. lines like `20:46:10,657 |-INFO in ch.qos.logback.classic.LoggerContext[default] - ...`

server/single-server/src/main/resources/application.conf
dpetrick: It influences the failover logic, leading to misleading error messages locally with the newest JDBC driver. Aurora has specifics for read replicas and failovers, and we don't need that locally.

server/single-server/src/main/scala/cool/graph/singleserver/SingleServerDependencies.scala
dpetrick: True, missed that.

PR #1121, opened 2017-11-02
[WIP] Adjust case for for operations in CLI messages

do4gr:
*lowercase operation spelling in CLI messages to match the expected input case


do4gr:
We only accept the operations in the yaml file if they are lowercased i.e. : Model.read but we were returning them as Model.Read as name of the model permission and as 'READ' in the CLI messages. Now it is uniformly in the case we accept as input.

PR #1153, opened 2017-11-03
allow function to return anything, even nothing in local mode

antho1404:
In local execution, the function called needed to return something. If the function didn't return anything we had a `Cannot read property 'then' of undefined`.

Like that now we can handle properly the functions that return nothing (really useful in subscription functions for exemple)


timsuchanek:
Thanks for the PR! This makes a lot of sense

PR #1167, opened 2017-11-05
Fix full example signup error caused by no name provided

oiojin831:


examples/full-example/src/email-password/signup.ts
```diff
@@ -82,6 +82,7 @@ async function createGraphcoolUser(api: GraphQLClient, email: string, password:
   const variables = {
     email,
     password: password,
+    name: name,
```
kbrandwijk: Only this one got left behind.

oiojin831: @kbrandwijk sorry :(
I just changed it.

marktani:
@oiojin831 can you elaborate what the problem is you're trying to fix? would @kbrandwijk's suggestion alleviate this problem?

kbrandwijk:
I think it just comes down to making `name` optional (`String!` -> `String`) in this full-example.

oiojin831:
@marktani 
@kbrandwijk is pointing out correctly.
Should I just change ```name``` to be optional?

nikolasburk:
@oiojin831 I think making `name` optional is the proper way here - I'll merge the PR once that's done. Thanks!

PR #1179, opened 2017-11-06
[WIP] Rabbit-backed tests to in-memory-backed tests

dpetrick:
Context: Having RabbitMQ in the tests is heavyweight and sometimes hard to handle correctly, even with the testkits. Additionally, we do not need to test the message bus RabbitMQ adapter in the tests, as it is sufficiently tested in the message bus rabbit tests - we only want to test the business logic.

Solution: Use lightweight, actor-backed in-memory testkits for the message bus interfaces for test speedup and robustness.

Work items:
- [x] More code documentation for testkits.
- [x] Finish implementation of in-memory messagebus testkits.
- [x] Add missing tests for in-memory queue testkit.
- [ ] Add missing tests for in-memory pubsub testkit.
- [ ] Replace rabbit testkits with in-memory testkits.
- [ ] Optional: Consolidate various testkit interfaces into a trait.

Additional notes:
- Various refactorings for testkits included to make the interface more uniform.
- The in-memory interface semantics differ slightly from the older rabbit interfaces: Expecting a message now reasons over the messages actually received by a subscriber / consumer, as opposed to the `-published` counterpart, which reasons over the incoming messages into the pubsub / queue. The old testkits did not distinguish those cases.

### DIFF
--- a/server/libs/message-bus/src/main/scala/cool/graph/messagebus/testkits/InMemoryPubSubTestKit.scala
+++ b/server/libs/message-bus/src/main/scala/cool/graph/messagebus/testkits/InMemoryPubSubTestKit.scala
@@ -5,17 +5,23 @@ import akka.stream.ActorMaterializer
 import akka.testkit.TestProbe
 import cool.graph.messagebus.Conversions.Converter
 import cool.graph.messagebus.PubSub
+import cool.graph.messagebus.pubsub._
 import cool.graph.messagebus.pubsub.inmemory.InMemoryAkkaPubSub
-import cool.graph.messagebus.pubsub.{Message, Only, Subscription, Topic}
 
-import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.language.existentials
 import scala.reflect.ClassTag
 
 /**
-  * InMemory testkit for simple test cases that requires reasoning over published or received messages.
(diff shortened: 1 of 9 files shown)


server/libs/message-bus/src/test/scala/cool/graph/messagebus/pubsub/inmemory/BasicPerformanceTesting.scala
```diff
@@ -0,0 +1,104 @@
...
+import akka.stream.ActorMaterializer
+import cool.graph.messagebus.pubsub.PubSubProtocol.Publish
+import cool.graph.messagebus.pubsub.{IntermediateCallbackActor, Message, PubSubRouterAlt, PubSubRouter}
+import org.scalatest.WordSpecLike
+
+import scala.reflect.ClassTag
+
+class BasicPerformanceTesting extends WordSpecLike {
+
+//  "Akka Routing implementation" should {
```
sorenbs: Do we need this?

dpetrick: Yes I'd keep it for future reference for now.

PR #1182, opened 2017-11-06
Add type to invalid name error message.

do4gr:
Fix for https://github.com/graphcool/framework/issues/994

The invalid name error was very generic and did not mention the type. It will now specify whether it was a model, field, relation or enum that had a naming convention violation.

Since several mutactions can throw the same naming violation error in one deploy we now also filter the error messages for duplicates.

### DIFF
--- a/server/backend-api-system/src/main/scala/cool/graph/system/mutations/PushMutation.scala
+++ b/server/backend-api-system/src/main/scala/cool/graph/system/mutations/PushMutation.scala
@@ -74,7 +74,7 @@ case class PushMutation(
 
   override def verifyActions(): Future[List[Try[MutactionVerificationSuccess]]] = {
     super.verifyActions().map { verifications =>
-      verifications.map {
+      val verificationResult = verifications.map {
         case Failure(error: Throwable) =>
           errors = errors :+ extractErrors(error)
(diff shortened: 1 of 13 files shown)


server/backend-api-system/src/main/scala/cool/graph/system/mutations/PushMutation.scala
```diff
@@ -51,7 +51,7 @@ case class PushMutation(
         errors = List(
           SchemaError(
             "Global",
-            "Only projects that have been ejected can make use of the CLI's deploy function. More details: https://docs-next.graph.cool/reference/service-definition/legacy-console-projects-aemieb1aev"
+            "Only projects that have been ejected can make use of the CLI's deploy function. More details: https://www.graph.cool/docs/reference/service-definition/legacy-console-projects-aemieb1aev/"
```
kbrandwijk: Don't update the link to the old docs. The correct link is: https://docs-next.graph.cool/docs/reference/service-definition/legacy-console-projects-aemieb1aev

kbrandwijk: Never mind, just noticed that those are actually the new docs...

Issue #1211, opened 2017-11-08
GRAPHCOOL_TARGET environment variable is not used as target in some CLI commands


vtsatskin:
**Current behavior**

Deploying to a graphcool target works without a `.graphcoolrc` file by setting the environment variable:

```
export GRAPHCOOL_TARGET=local/cj9evgxgm00040174yoi292dw
graphcool deploy
```

However, this is not the case for the following commands which also support a target parameter:

* `graphcool playground`
* `graphcool info`
* `graphcool root-token`
* `graphcool logs`
* `graphcool delete`
* `graphcool invoke-local`

**Reproduction**

1) Delete the `.graphcoolrc` file.
2) Set GRAPHCOOL_TARGET environment variable to a valid target
3) Use one of the aforementioned commands
4) See error:
> Please provide a valid target that points to a valid cluster and service id

**Expected behavior?**

The aforementioned commands respect the `GRAPHCOOL_TARGET` environment variable.

**Discussion**

If the expected behaviour is not desired and this isn't a bug, perhaps I am misunderstanding the intention of the existence of `GRAPHCOOL_TARGET`. I will explain how I use the environment variable in a multiple developer team. I would be interested in knowing how the CLI is designed to be used in my situation.

My team runs a local graphcool cluster when developing locally and deploys to production when the development work is complete. So being able to share production deployment information is important, which is what I believe the intent of the `.graphcoolrc` file is. I assume that this file is intended to be checked into version control and shared with team members.

Using a `.graphcoolrc` file in version control is not optimal in multiple developer projects due to the local target id changing for each developer when they deploy locally. This results in changes to the  `.graphcoolrc` file. However, sharing a common target for production deployment is desirable. The way I've gotten around this issue is to not use a `.graphcoolrc` file, delete it from source control, store environment variables in a file, and load those in a script depending on which environment is desired.

Two environment configuration files are checked into source control:

**.env.production**
```
GRAPHQL_API_ENDPOINT=https://api.graph.cool/simple/v1/<production id> 
GRAPHCOOL_TARGET=<cluster id>/<production id>
```

**.env** (Local, intentionally blank)
```
GRAPHCOOL_TARGET=
GRAPHQL_API_ENDPOINT=
```

While another file is not checked in and used for local configuration by each team member:

**.env.local**
```
GRAPHCOOL_TARGET=local/<local id>
GRAPHQL_API_ENDPOINT=http://localhost:60000/simple/v1/<local id>
```

[...]

PR #1212, opened 2017-11-08
Added target env var to all commands

kbrandwijk:
Proposal for #1211

### DIFF
--- a/cli/packages/graphcool-cli-core/src/commands/console/index.ts
+++ b/cli/packages/graphcool-cli-core/src/commands/console/index.ts
@@ -16,7 +16,7 @@ export default class Console extends Command {
     await this.auth.ensureAuth()
     let {target} = this.flags
 
-    const foundTarget = await this.env.getTarget(target)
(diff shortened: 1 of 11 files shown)


kbrandwijk:
I agree this could be more standardized. I'm just wondering how exactly. There's actually another issue with this. The env var overrides the passed in target now, it should be the other way around I think:

`this.flags.target || process.env.GRAPHCOOL_TARGET || defaultTarget`

Do you agree with that? 

I also like to do this in once place. With the `flags.target` being leading, I agree that this can move to `getTarget`. If you agree with the above, I'll make the changes.

timsuchanek:
Ok that makes a lot of sense! Yes the CLI arg should overwrite the env var. Ok then I'm looking forward for the changes :)

timsuchanek:
Sorry for not moving foward with the PR for so long, the code for the framework has been moved to https://github.com/graphcool/graphcool-framework.
Any further PRs can be done to that repository.

Issue #1008, opened 2017-10-23
graphcool delete fails locally


timwis:
I hope this is helpful but I apologise if it lacks details or is difficult to reproduce. I ran into this while troubleshooting and wanting to start over without leaving 4 docker machines running in the background.

**Current behavior**
```bash
graphcool delete
? Are you sure that you want to delete cj943qp0l000601098v37d5do? y/N y
Deleting project cj943qp0l000601098v37d5do... !

ERROR: Cannot query field 'deleteProject' on type 'Mutation'. Did you mean 'addProject'? (line 3, column 7):
      cj943qp0l000601098v37d5do: deleteProject(input:{
      ^

{
  "data": null,
  "errors": [
    {
      "message": "Cannot query field 'deleteProject' on type 'Mutation'. Did you mean 'addProject'? (line 3, column 7):\n      cj943qp0l000601098v37d5do: deleteProject(input:{\n      ^",
      "locations": [
        {
          "line": 3,
          "column": 7
        }
      ]
    }
  ],
  "status": 400
}
```

timwis:
Same as #964

marktani:
For now, you can remove the docker containers, **which will delete all local Graphcool services irreversibly**.

With `docker ps` you get the ids of the following containers:

* `graphcool/graphcool-dev`
* `mysql`
* `graphcool/localfaas`
* `rabbitmq`

Then kill & delete them:

```sh
docker kill <container id>
docker rm <container id>
```

timwis:
Thanks -- is there, by any chance, a mounted volume that's persisting the data/configuration of those images? Because I did that, but when I try to reinitialise, I'm told that a project with that name already exists in my account (on my local cluster).

```bash
? Please choose the cluster you want to deploy to

local

? Please choose the target name dev
Creating service textblast locally in cluster local... !

ERROR: A project with the name 'textblast' already exists in your account

{
  "data": {
    "addProject": null
  },
  "errors": [
    {
      "locations": [
        {
          "line": 2,
          "column": 9
        }
      ],
      "path": [
        "addProject"
      ],
      "code": 2003,
      "message": "A project with the name 'textblast' already exists in your account",
      "requestId": "eu-west-1:system:cj94sd7eh000a0172vspxze7d"
    }
  ],
  "status": 200
}

Get in touch if you need help: https://www.graph.cool/forum
To get more detailed output, run $ export DEBUG="*"
```

Toosick:
[...]

PR #1213, opened 2017-11-08
Added guard for delete local similar to console local

kbrandwijk:
As a temporary user-friendly fix for #1008


cli/packages/graphcool-cli-core/src/commands/delete/index.ts
```diff
@@ -31,15 +31,20 @@ export default class Delete extends Command {
     const foundTarget = await this.env.getTargetWithName(target)
 
     if (foundTarget && foundTarget.target) {
-      const id = foundTarget.target.id
-      if (!force) {
-        await this.askForConfirmation(id)
+      if (!this.env.isSharedCluster(foundTarget.cluster)) {
+        this.out.error(`Can't delete service in local cluster ${foundTarget.cluster}.
+This command only available in the hosted version of Graphcool.`)
```
marktani: > This command is only

PR #1219, opened 2017-11-09
Warn for any existing files on init

### LINKED ISSUE #1103: graphcool init overwrites existing package.json
_From @wesbos on October 27, 2017 14:7_

running `graphcool init` just blew away my existing package.json - it should check if there is already a package.json 

_Copied from original issue: graphcool/graphql-cli#41_

kbrandwijk:
This PR depends on:
- [ ] `proposal/accepted` on #1103

Fixes #1103. The current implementation only warns if there is a `graphcool.yml` file in the folder when running `gc init`. I extended that check to warn on any files in the folder. This fixes the issue with overwriting package.json, but better yet, it aligns with the recommendation to always create your Graphcool project in an empty folder. The warning message is also updated accordingly.

### DIFF
--- a/cli/packages/graphcool-cli-core/src/commands/init/index.ts
+++ b/cli/packages/graphcool-cli-core/src/commands/init/index.ts
@@ -64,15 +64,13 @@ export default class Init extends Command {
       this.out.log(`
 The directory ${chalk.green(
         this.config.definitionDir,
-      )} contains files that could conflict:
+      )} is not empty.
 
-${files.map(f => `  ${f}`).join('\n')}
-
-Either try using a new directory name, or remove the files listed above.
+It is recommended to initialize your service in a new directory.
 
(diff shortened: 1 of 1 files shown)


timsuchanek:
The first implementation looked for any file, too.
The problem with this is, for example, the workflow with a new Serverless project.
Then you execute
```
sls create --template template
gc init
-> error
```
That's not a nice getting started experience. So instead, it should warn, "Hey, there is already a package.json. We don't override it, but please keep it in mind"

kbrandwijk:
You don't want your serverless project in the same folder either.

timsuchanek:
Sorry for not moving foward with the PR for so long, the code for the framework has been moved to https://github.com/graphcool/graphcool-framework.
Any further graphcool framework related PRs can be done to that repository.

PR #1247, opened 2017-11-11
docs/auth: Remove --token flag

timwis:
This flag doesn't appear to exist anymore

### DIFF
--- a/docs/04-Reference/06-Auth/02-Authentication/02-Authentication-Tokens.md
+++ b/docs/04-Reference/06-Auth/02-Authentication/02-Authentication-Tokens.md
@@ -120,10 +120,8 @@ When your service is deployed, the corresponding [target](!alias-zoug8seen4#mana
 You can obtain the value of the root token using the [`graphcool root-token`](!alias-aiteerae6l#graphcool-root-token) command:
 
 ```sh
-graphcool root-token --token myToken1
+graphcool root-token myToken1
 ```
-
-If you don't pass the `--token` option to the command, it will simply print the names of all the root tokens associated with this target.
 
 
 ### Using a _temporary_ root token inside functions


docs/04-Reference/06-Auth/02-Authentication/02-Authentication-Tokens.md
```diff
@@ -120,11 +120,9 @@ When your service is deployed, the corresponding [target](!alias-zoug8seen4#mana
 You can obtain the value of the root token using the [`graphcool root-token`](!alias-aiteerae6l#graphcool-root-token) command:
 
 ```sh
-graphcool root-token --token myToken1
+graphcool root-token myToken1
 ```
```
kbrandwijk: I would keep this part, but saying:
`If you don't pass the token name to the command, it will simply print the names of all the root tokens associated with this target.`

timwis: @kbrandwijk Actually, if you exclude `--token`, you get an error, `Unknown flag --token`

kbrandwijk: I was referring to the part you removed about not passing a --token. That still works the same, but not with the --token flag, but with the token name itself. In order words, if you don't pass a _token name_, you get all the tokens.

timwis: Oh! That makes sense. I'll try to edit that tomorrow.

nikolasburk:
Thanks @timwis and @kbrandwijk! I'm adding the hint that the command prints token names when no token name is provided as an argument myself 👍

PR #1248, opened 2017-11-11
[WIP] Added validation for event payload

### LINKED ISSUE #1240: Create JSON schema for event payload and add validation for `graphcool invoke-local`
**What feature are you missing?**
You can currently throw any JSON file at `graphcool invoke-local`. However, if the structure was wrong, it will cause runtime error down the line, when the deployed function is called with the actual event argument.

It would be helpful to check the structure of the JSON to see if it's valid.

**How could this feature look like in detail? Tradeoffs?**
Check for `event.data`, `event.context` and the subfields of `event.context`. Better yet, create a JSON schema for the this, and just validate it. Preferably using my favorite validator: https://github.com/epoberezkin/ajv

kbrandwijk:
Do not merge directly. This PR depends on:

- [ ] PR for graphcool-json-schema: graphcool/graphcool-json-schema#6
- [ ] Release new `graphcool-json-schema` version and update dependencies in CLI packages
- [ ] `proposal/accepted` on #1240

This adds validation of the event payload for `gc invoke-local` using the new schema in `graphcool-json-schema`.

Fixes #1240

### DIFF
--- a/cli/packages/graphcool-cli-core/src/commands/invoke/local.ts
+++ b/cli/packages/graphcool-cli-core/src/commands/invoke/local.ts
@@ -1,4 +1,6 @@
 import { Command, flags, Flags, GraphcoolModule } from 'graphcool-cli-engine'
+import * as Ajv from 'ajv'
+import * as schema from 'graphcool-json-schema/dist/eventschema.json'
 import { flatMap, isEqual } from 'lodash'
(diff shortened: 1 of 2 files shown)


timsuchanek:
This has already been implemented in the graphcool-framework cli.

PR #1252, opened 2017-11-12
List all local clusters on interactive deploy

### LINKED ISSUE #905: Cannot deploy interactively to second local cluster
**Current behavior**
I can't select a second local cluster with `graphcool deploy`

**Reproduction**
- I create a second local cluster with `graphcool local up --name second`.
- The cluster is succesfully added to ~/.graphcoolrc
- When I run `graphcool deploy -t newstage`, It only shows one local cluster (local).

**Expected behavior?**
All local clusters show up.

kbrandwijk:
Fixes #905

### DIFF
--- a/cli/packages/graphcool-cli-core/src/commands/deploy/index.ts
+++ b/cli/packages/graphcool-cli-core/src/commands/deploy/index.ts
@@ -359,6 +359,9 @@ https://console.graph.cool/${encodeURIComponent(info.name)}/settings/general`)
   }
 
   private async clusterSelection(): Promise<string> {
+    let localClusters = Object.keys((this.env.rc || {}).clusters as {} || {})
+    if (!localClusters.length) localClusters = ['local']
+
     const question = {
       name: 'cluster',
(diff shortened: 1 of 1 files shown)


cli/packages/graphcool-cli-core/src/commands/deploy/index.ts
```diff
@@ -359,6 +359,9 @@ https://console.graph.cool/${encodeURIComponent(info.name)}/settings/general`)
   }
 
   private async clusterSelection(): Promise<string> {
+    let localClusters = Object.keys((this.env.rc || {}).clusters as {} || {})
```
timsuchanek: Proposal for these 2 lines:
```js
const localClusters = (this.env.rc && this.env.rc.clusters) ?
                      Object.keys(this.env.rc.clusters) : ['local']
```

From our current code style, we don't use `if` statements without curly braces. We should specify the style guide.

timsuchanek:
Closing this due to inactivity.

PR #1254, opened 2017-11-12
No online authorization required for local cluster commands

### LINKED ISSUE #1067: `graphcool logs` asks for authentication on local target
**Current behavior**
I have set up a service, running locally, and I want to access the logs. When I run `gc logs`, I need to authenticate online, which is totally unnecessary.

**Expected behavior?**
Show the local logs without authentication.

kbrandwijk:
Fixes #1067.

Applies to `gc logs`, `gc info` and `gc root-token`.


timsuchanek:
Thanks for the PR!

PR #1266, opened 2017-11-13
[WIP] Detect invalid directives in the schema

do4gr:
covers https://github.com/graphcool/framework/issues/992


kbrandwijk:
I like the fact that it validates known directives, but it doesn't take into account the support for adding directives that are ignored by Graphcool, but needed in (for example) the Gateway.
I also think this change makes it harder to support that in a next stage.

do4gr:
Thanks for raising these issues Kim, we'll discuss this internally before proceeding.

marktani:
We'll revisit this at a later point.

PR #1289, opened 2017-11-17
fix typo

michaelchiche:


nikolasburk:
Thanks so much for catching this @michaelchiche and sorry for the delay 🙏

PR #1294, opened 2017-11-18
fix example which can't work on authenticate react tutorial

joe-re:
I found a small mistake in docs.


nikolasburk:
Thanks a lot @joe-re 🎉

PR #1326, opened 2017-11-26
fix(docs): GenerateRootToken should be GenerateNodeToken

picosam:


nikolasburk:
Thanks a lot @picosam 🎉

PR #1344, opened 2017-11-29
fix bad mutation example

michaelchiche:


nikolasburk:
Thanks again @michaelchiche 🎉

PR #1360, opened 2017-12-04
Update 01-Node.md

chaim1986:
To avoid the error  when query from API gateway...

---- ERROR ----
[nodemon] starting `node index.js`
Server running. Open http://localhost:3000/playground to run queries.
Error: Schema must be an instance of GraphQLSchema. Also ensure that there are not multiple versions of GraphQL installed in your node_modules directory...."

### DIFF
--- a/docs/01-Quickstart/02-Backend/02-Node/01-Node.md
+++ b/docs/01-Quickstart/02-Backend/02-Node/01-Node.md
@@ -212,6 +212,7 @@ Navigate to the `gateway` directory and start the server:
 
 ```bash(path="gateway")
 cd ../gateway
+yarn add graphql
 yarn install
 yarn start
 ```


nikolasburk:
Hi @chaim1986! Thanks a lot for the PR, I'm not sure why you were running into this error. I just went through all the quickstart instructions and it does work for me. Also, `graphql` already is [listed a dependency](https://github.com/graphcool/framework/blob/master/examples/node-gateway-custom-schema/gateway/package.json#L11) of the gateway app, so adding it once more before running `yarn start` doesn't actually change anything. I'm closing this PR, please feel free to [open an issue](https://github.com/graphcool/framework/issues/new) with a bit more info about your setup if you're still experiencing this problem!

PR #1371, opened 2017-12-07
Fix docs: rename generateAuthToken to generateNodeToken

danistefanovic:
If I understand it correctly, [is `generateAuthToken` deprecated](https://github.com/graphcool/framework/issues/874).

(There is also an open [PR in the graphcool-lib](https://github.com/graphcool/graphcool-lib/pull/15) repo that fixes the example in the graphcool-lib README)


nikolasburk:
Thanks a lot for catching this one @danistefanovic 🙏

PR #1391, opened 2017-12-13
fixed print-out typo

xeeo:
when invoking a local function the CLI suggests an incorrect command: "invoke local", instead of "invoke-local"


timsuchanek:
Thanks for the PR!

PR #1459, opened 2017-12-29
Handle for a missing .graphcool directory

### LINKED ISSUE #1451: npm install -g graphcool does not work on windows with latest release
After running `npm install -g graphcool `, when i try to run `gcf`, i get the following message:
```

 !    'ENOENT': ENOENT: no such file or directory, lstat 'C:\Users\Administrador\.graphcool'

Get in touch if you need help: https://www.graph.cool/forum
To get more detailed output, run $ export DEBUG="*"
```

The previous version `0.10.1` seems to work ok.

FredyC:
Fixes #1451

The `lstat` throws an exception when trying to access no existing file/directory. Another way would be `try/catch` block, but since there is already check for `exists` then this is enough.


timsuchanek:
Thanks for the PR!

PR #1507, opened 2018-01-04
[WIP] Examples 1.0

marktani:
Hold off merging until later 🙂


examples/passthrough/README.md
```diff
@@ -27,3 +27,7 @@ mutation c {
   }
 }
 ```
+
+### Questions
+
+why is it called `db` on `forwardTo(`'db')`: because it is called like this in index.js
```
kbrandwijk: Because that's the name of the binding in index.js

PR #1545, opened 2018-01-08
Add missing subject in deploying services doc

wacii:
Just fixing a small mistake I found while reading through the docs.


marktani:
Thanks 🙌

PR #1575, opened 2018-01-11
Fix graphcool login --token

mattfysh:


timsuchanek:
Thanks for the PR, but unfortunately, I can't merge it here. The `graphcool-framework` code has moved to https://github.com/graphcool/graphcool-framework.

This issue is fixed in the `prisma` CLI. There it is now called `prisma login --key MY_KEY`.
If you need this functionality in the graphcool cli, please do a PR there.

PR #1615, opened 2018-01-13
Removing local cluster from graphcool.yml

qucode1:
- Remove local cluster option from 1.0 examples, to allow choosing a cluster on deployment.

- update download link in permissions example

### DIFF
--- a/examples/1.0/file-api/database/graphcool.yml
+++ b/examples/1.0/file-api/database/graphcool.yml
@@ -6,4 +6,3 @@ cluster: local
 datamodel:
   - datamodel.graphql
   
-secret: ${env:GRAPHCOOL_SECRET}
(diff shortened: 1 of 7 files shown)


examples/1.0/file-api/database/graphcool.yml
```diff
@@ -6,4 +6,3 @@ cluster: local
 datamodel:
   - datamodel.graphql
   
-secret: ${env:GRAPHCOOL_SECRET}
```
marktani: why did you remove the secret here?

qucode1: sorry that is not what I had in mind. I will fix that.

marktani:
Thanks a lot @qucode1 🎉

PR #1617, opened 2018-01-13
Update 05-File-Management Doc

joefazz:
When first using the File Management you will get a 500 error when there is no File type to be found, based off several forum posts and my own personal experience I think this addition will make things clearer for users. I'm personally using version 0.x, if this change is applicable to 1.0 I'd be happy to add it there also.


nikolasburk:
Hey @joefazz, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #1630, opened 2018-01-16
Updates File API Example

DevanB:
**UNTESTED**


lunchboxer:
now you're just duplicating what was here: https://github.com/maticzav/graphql-server-file-upload-example There was a nice, simple example to show how to upload a file to the server. I'd like to vote that we keep the old one and improve it. This AWS upload example can live next to it.

PR #1636, opened 2018-01-16
fix type in node token with graphcool-lib example

mikhailvs:
`const graphcool-framework = ...` is not valid code. I believe that is meant to say `const graphcool = ...`


nikolasburk:
Hey @mikhailvs, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #1638, opened 2018-01-16
fix type description to Json from JSON on docs

joe-re:
Graphcool provides Json type is Json, not JSON.
It causes confused usage.

I think it is better to fix it as match actually data type.


marktani:
Thanks, I applied this to the `graphcool-framework` docs here: https://github.com/graphcool/graphcool-framework/commit/b4f2422d99341ed22e7c21d6f17172b070c644e7

PR #1668, opened 2018-01-17
We should only use @relation when needed.

mavilein:


nikolasburk:
Thanks for the catch @mavilein!

PR #1689, opened 2018-01-19
Make sure to delete RelayIds when deleting nodes

do4gr:
- fixes it for nested deletes and deleteMany
- also cleans up DatabaseMutationbuilder
- also removes the remains of FieldMirrorColumns from the code


server/api/src/main/scala/com/prisma/api/database/DatabaseApiTestDatabaseMutationBuilder.scala
```diff
@@ -0,0 +1,123 @@
+package com.prisma.api.database
+
+import com.prisma.shared.models.TypeIdentifier.TypeIdentifier
+import com.prisma.shared.models.{Model, TypeIdentifier}
+import slick.dbio.DBIOAction
+import slick.jdbc.MySQLProfile.api._
+
+object DatabaseApiTestDatabaseMutationBuilder {
```
mavilein: If this is indeed just used in tests, it should reside in `src/test/scala`

server/api/src/main/scala/com/prisma/api/database/mutactions/mutactions/DeleteDataItemNested.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.prisma.api.mutations.NodeSelector
+import com.prisma.api.schema.APIErrors
+import com.prisma.shared.models.Project
+import slick.jdbc.MySQLProfile.api._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
+
+case class DeleteDataItemNested(project: Project, where: NodeSelector) extends ClientSqlDataChangeMutaction {
```
mavilein: Could we directly substitute the uses of `DeleteDataItem` with this one? This one looks better to me.

server/api/src/main/scala/com/prisma/api/database/DatabaseApiTestDatabaseMutationBuilder.scala
do4gr: moved it

server/api/src/main/scala/com/prisma/api/database/mutactions/mutactions/DeleteDataItemNested.scala
do4gr: We can probably do that at a later point when we do not discern between top-level and nested mutactions that heavily anymore. At the moment that change would require a lot of extra refactoring.

PR #1694, opened 2018-01-19
Update links to graphql-yoga examples

ericapisani:
I found a few links that were out of date and updated them to go to the new locations.


nikolasburk:
Thanks so much for catching & fixing these @ericapisani!! 🎉

PR #1707, opened 2018-01-21
Enable Scalar Lists in Upsert

do4gr:
- nested mutations in upserts are still not enabled, this will follow in a later PR

### DIFF
--- a/server/api/src/main/scala/com/prisma/api/mutations/SqlMutactions.scala
+++ b/server/api/src/main/scala/com/prisma/api/mutations/SqlMutactions.scala
@@ -1,7 +1,7 @@
 package com.prisma.api.mutations
 import com.prisma.api.database.mutactions.ClientSqlMutaction
 import com.prisma.api.database.mutactions.mutactions._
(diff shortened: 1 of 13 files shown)


Huvik:
❤️

server/api/src/main/scala/com/prisma/api/mutations/SqlMutactions.scala
```diff
@@ -121,6 +127,18 @@ case class SqlMutactions(dataResolver: DataResolver) {
     x.flatten.toVector
   }
 
+  def getMutactionsForScalarLists2(where: NodeSelector, args: CoolArgs) = {
```
mavilein: do we still need the old `getMutactionsForScalarLists `?

do4gr: Yes, now one delivers Mutactions and the other DBActions. I already cleaned up the naming in the follow up. And maybe can unite them again once we have the control flow abstraction. But for now they do different things.

PR #1747, opened 2018-01-26
Updates Examples in README

DevanB:
This PR:
- Removes `rest-wrapper`, which is no longer in the examples directory
- Corrects `file-api` to `file-handling-s3`


marktani:
thanks 👌

PR #1777, opened 2018-01-30
Fixed post resolver in permission example

jpaquim:
Added missing userId definition.


nikolasburk:
Thanks so much for catching this @jpaquim 🎉

PR #1792, opened 2018-01-31
Remove required from author property

idkjs:
Since the author property is required, the subsequent import fails since the test data createPost doesnt have an author property on it. Works if you take off required !


johannpinson:
Not sure that deleting the "required" flag is the best option.
The ideal way is to fix the import function/command directly, check https://github.com/graphcool/prisma/issues/1767 😉

PR #1811, opened 2018-02-02
docs: fix wrong query argument

akoenig:
This is just a documentation fix. The headline says `Retrieve a single node by its email`, but the example passes an `id` as the argument.


nikolasburk:
Thanks a lot @akoenig 🙌

PR #1812, opened 2018-02-02
fix mysql password typo

jamesmbowler:
I wasn't able to log into my local docker container msyql until I searched the code and found the correct password.


marktani:
Thanks!

PR #1826, opened 2018-02-05
docs proposal: switch to bool default value

akoenig:
Even if both is working, I think that defining the actual `bool` default value NOT as a string might be less confusing. I just stumbled upon this one and thought: "Ha, why is that a 'string'?". Just my opinion and I thought that you might be interested in my thought process :) Feel free to close this PR if you want to stick with the string definition :)

### DIFF
--- a/docs/1.1/04-Reference/01-Service-Configuration/03-Data-Modelling-(SDL).md
+++ b/docs/1.1/04-Reference/01-Service-Configuration/03-Data-Modelling-(SDL).md
@@ -347,7 +347,7 @@ To specify a default value for a field, you can use the `@default` directive:
 
 ```graphql
 type Story {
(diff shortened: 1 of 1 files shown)


akoenig:
haha .. just read further and noticed the remark in the next paragraph: `Notice that you need to always provide the value in double-quotes, even for non-string types such as Boolean or Int.`

marktani:
That's correct 🙂 Thanks for your feedback anyway!

nagman:
So why is it working with non-string values?

I mean that this works even without double quotes:
```
type Foo {
  active: Boolean @default(value: false)
}
```
Is there a reason why we should use quotes?  
As @akoenig, I think it would be less confusing without.

PR #1830, opened 2018-02-06
Fix typos

ciwchris:


nikolasburk:
Thanks so much for catch all of these @ciwchris 🔥🔥🔥🔥

PR #1832, opened 2018-02-06
Backwards compatible formatter for relations

do4gr:


server/deploy/src/main/scala/com/prisma/deploy/migration/inference/SchemaInferrer.scala
```diff
@@ -161,20 +177,58 @@ case class SchemaInferrerImpl(
...
+            case Some(otherDirective) => directive.argument_!("name").value.renderCompact == otherDirective.argument_!("name").value.renderCompact
+            case None                 => false
+        })
+      case None => otherFieldsOnModelBRelatedToModelA.headOption
+    }
+    field
+  }
+
+  private def getOnDeleteFromField(field: FieldDefinition): OnDelete.Value = {
+    field.directive("relation") match {
```
mavilein: you can remove the double nested matches if chain via flatMap:
```
field.directive("relation").flatMap(_.argument("onDelete"") match ...
```
But i think you could just use our existing `directiveArgumentAsString` [method](https://github.com/graphcool/prisma/blob/7b3da2706b9c80a4f17bfa2e571c11618d58f735/server/deploy/src/main/scala/com/prisma/deploy/migration/DataSchemaAstExtensions.scala#L114).

server/shared-models/src/main/scala/com/prisma/shared/models/ProjectJsonFormatter.scala
```diff
@@ -163,7 +165,38 @@ object ProjectJsonFormatter {
     private def addDiscriminator(json: JsObject, fn: Function) = json ++ Json.obj(discriminatorField -> fn.typeCode.toString)
   }
 
-  implicit lazy val relation                  = Json.format[Relation]
+  implicit val relationWrites: Writes[Relation] = (
```
mavilein: I think we should remove the implicit keyword for the Writes and Reads. It is enough that the Format is marked implicit.

server/shared-models/src/main/scala/com/prisma/shared/models/ProjectJsonFormatter.scala
```diff
@@ -163,7 +165,38 @@ object ProjectJsonFormatter {
...
+      (JsPath \ "name").read[String] and
+        (JsPath \ "description").readNullable[String] and
+        (JsPath \ "modelAId").read[String] and
+        (JsPath \ "modelBId").read[String] and
+        (JsPath \ "modelAOnDelete").readNullable[OnDelete.Value] and
+        (JsPath \ "modelBOnDelete").readNullable[OnDelete.Value]
+    )
+
+    x.apply { (name, description, modelAId, modelBId, modelAOnDelete, modelBOnDelete) =>
+      Relation(
```
mavilein: While reading the code i realised that we do not need to this explicit invocation of the constructor.
Above we can just do:
```
(JsPath \ "modelAOnDelete").readNullable[OnDelete.Value].map(_.getOrElse(OnDelete.SetNull)) and 
...
```
Therefore we can use the shorter magic syntax again like in:
```
implicit val placeReads: Reads[Place] = (
  (JsPath \ "name").read[String] and
  (JsPath \ "location").read[Location] and
  (JsPath \ "residents").read[Seq[Resident]]
)(Place.apply _)
```

server/deploy/src/main/scala/com/prisma/deploy/migration/inference/SchemaInferrer.scala
do4gr: Done.

server/shared-models/src/main/scala/com/prisma/shared/models/ProjectJsonFormatter.scala
do4gr: Done

server/shared-models/src/main/scala/com/prisma/shared/models/ProjectJsonFormatter.scala
do4gr: Done

PR #1876, opened 2018-02-12
Fixed error in readme (missing array for nested inserts)

ntziolis:


nikolasburk:
Thanks a lot @ntziolis 🎉

PR #1878, opened 2018-02-12
Update and Upsert missing in nested API

do4gr:
Fix for https://github.com/graphcool/prisma/issues/1837
- generate the update and upsert input types when there is no backrelation
- adds relation handling for the create case of upsert
- also adds the relay id table entries in the create case of upsert
- add relation test cases for nested upsert in update

### DIFF
--- a/server/api/src/main/scala/com/prisma/api/database/DatabaseMutationBuilder.scala
+++ b/server/api/src/main/scala/com/prisma/api/database/DatabaseMutationBuilder.scala
@@ -3,6 +3,7 @@ package com.prisma.api.database
 import com.prisma.api.mutations.mutations.CascadingDeletes.Path
 import com.prisma.api.database.SlickExtensions._
 import com.prisma.api.database.Types.DataItemFilterCollection
+import com.prisma.api.database.mutactions.mutactions.NestedCreateRelationMutaction
 import com.prisma.api.mutations.{CoolArgs, NodeSelector, ParentInfo}
 import com.prisma.api.schema.GeneralError
 import com.prisma.shared.models.TypeIdentifier.TypeIdentifier
@@ -12,6 +13,7 @@ import slick.dbio.{DBIOAction, Effect, NoStream}
 import slick.jdbc.MySQLProfile.api._
 import slick.jdbc.SQLActionBuilder
 import slick.sql.{SqlAction, SqlStreamingAction}
+
 import scala.concurrent.ExecutionContext.Implicits.global
 
 object DatabaseMutationBuilder {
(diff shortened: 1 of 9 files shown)


server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+        |
+      """.stripMargin
+    }
+    database.setup(project)
+
+    createItem(project, "Cat", "garfield")
+    createItem(project, "Owner", "jon")
+
+    //set initial owner
+    val res = server.executeQuerySimple(
```
mavilein: Couldn't we do this setup block also as one nested mutation? This way it is a bit hard to read.

server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+      """mutation {updateOwner(where: {ownerName: "jon"},
+        |data: {cat: {update: { where:{catName: "garfield"}, data: {catName: "azrael"}}}}) {
+        |    ownerName
+        |    cat {
+        |      catName
+        |    }
+        |  }
+        |}""".stripMargin,
+      project
+    )
```
mavilein: why isn't the result checked here? 🤔

server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  "Nested Upsert" should "be generated models with missing backrelations " in {
```
mavilein: This test is rather checking the execution of the mutation instead of the generated schema, right?

server/api/src/main/scala/com/prisma/api/database/DatabaseMutationBuilder.scala
```diff
@@ -287,6 +295,20 @@ object DatabaseMutationBuilder {
...
 
+  def ifThenElseNestedUpsert(condition: SqlStreamingAction[Vector[Boolean], Boolean, Effect],
+                             thenMutactions: DBIOAction[Unit, NoStream, Effect],
+                             elseMutactions: DBIOAction[Unit, NoStream, Effect],
+                             elseChecks: NestedCreateRelationMutaction) = {
+    import scala.concurrent.ExecutionContext.Implicits.global
+    for {
+      exists <- condition
+      action <- if (exists.head) { thenMutactions } else {
+                 val list = List(elseMutactions) ++ elseChecks.requiredCheck ++ elseChecks.removalActions ++ elseChecks.addAction
```
mavilein: the order of checks here is the same as in `.execute()` in the Mutaction. It might be a good idea to refactor those actions in a separate val, e.g.: `elseChecks.allActions`

mavilein: I have problems understanding the name `elseChecks`. It is also acting on the relation data itself. Should it maybe be named `relationMutactions`?

server/api/src/main/scala/com/prisma/api/database/DatabaseMutationBuilder.scala
```diff
@@ -297,26 +319,19 @@ object DatabaseMutationBuilder {
     } yield action
   }
 
-  def triggerFailureWhenNotExists(project: Project, query: SQLActionBuilder, table: String) = {
+  def triggerFailureWhenExists(project: Project, query: SQLActionBuilder, table: String, notExists: Boolean = false) = {
```
mavilein: We should try to avoid boolean in public signature. I think we should rather two method names, e.g. `triggerFailureWhenExists` and `triggerFailureWhenNotExists`.

mavilein: Ah, now i just saw exactly that below. I would do two things to avoid this confusion:
1. Move `triggerFailureWhenNotExists` to allow reading from top to bottom.
2. Add a private method `triggerFailureWhenInternal` that takes this boolean. So there will be 3 methods: `triggerFailureWhenNotExists`, `triggerFailureWhenExists`, `triggerFailureWhenInternal`. And only the private one contains the complex boolean.

do4gr: Done.

server/api/src/main/scala/com/prisma/api/database/DatabaseMutationBuilder.scala
do4gr: Done

server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
do4gr: Done

server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
do4gr: I only wanted to check that the mutation was now working and was relying on the must fail check.

server/api/src/test/scala/com/prisma/api/mutations/OptionalBackrelationSpec.scala
do4gr: Yes, i adjusted the title of the test.

PR #1893, opened 2018-02-13
Self relation problems in Nested Connect

do4gr:
Self-relations would sometimes error since relation.sideOf(model) returns wrong results for self relations.

Fix for: https://github.com/graphcool/prisma/issues/1609


akoenig:
Awesome, @do4gr. Thanks for fixing it! 🎉

PR #1911, opened 2018-02-15
UpdateMutation regression

do4gr:
- rename the InputTypes for models with optional backrelations so that they are unique
- verify schema generation for optional back relations

Fixes https://github.com/graphcool/prisma/issues/1909


sorenbs:
LGTM!

PR #1916, opened 2018-02-15
docs: added an example for subscriptions

Lauchi:
… better

### DIFF
--- a/docs/1.2/04-Reference/04-Server_side-Subscriptions/01-Overview.md
+++ b/docs/1.2/04-Reference/04-Server_side-Subscriptions/01-Overview.md
@@ -49,7 +49,35 @@ subscriptions:
       }
 ```
 
-### Example
+## Adding a Subscription as Resolver
+You can also add the Subscription directly as a resolver:
+
+```
+const resolvers = {
+...
+Subscription: {
+  userChangedEmail: {
+    subscribe: async (parent, args, ctx, info) => {
+      return await ctx.db.subscription.user(
+        {
+          where: {
+            mutation_in: ['UPDATED'],
+            updatedFields_contains: "email"}
+        }, info);
+    },
+  },
+},
+```
+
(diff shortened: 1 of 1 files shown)


nikolasburk:
Hey @Lauchi, thanks a lot for this PR! 🙏  

The page you submitted the PR for is about _server-side_ subscriptions which is a different feature than the subscription you added. However, it's clear to me that the docs do need a lot of improvement here! Currently, it's not really clear what a server-side subscriptions actually is and why it's different from the one you added (after all, you subscription is also happening on the server-side, so already the naming is pretty confusing...).

Server-side subscriptions specifically describe the mechanism to connect a subscription in the Prisma API with a _webhook_ (e.g. a lambda function) configured through your prisma.yml file. See the docs [here](https://www.prismagraphql.com/docs/reference/service-configuration/prisma.yml/yaml-structure-ufeshusai8#subscriptions-(optional)).

Still, thanks so much for your effort and please feel continue to open PRs in the future wherever you think our docs or examples might need improvement!! 🙏

marktani:
To add on this, what could be improved [in this example](https://github.com/graphcool/prisma/tree/master/examples/subscriptions) that would help you understand subscription resolvers better, @Lauchi?

Maybe we can turn your PR into an improvement to that example instead 🙂

Lauchi:
@marktani If there is a fundamental difference between a subscription and a server side subscription, I would have expected another Page with "Subscriptions" that would explain the example I gave. Until now at least I was not aware that there is a difference ;)

I was actually thinking that a server side subscription was something that i could use as some kind of an event store. Like listening on the "server side" to my own events and reacting accordingly. I also tought, that you had to add some kind of js function to the subscription to execute something like sending a mail. In the graphcool cloud you get a schema and a function if you create a server side subscription. I was expecting an explanation on how to do that when using prisma. So far I was not able to do that :/

PR #1931, opened 2018-02-17
Update index.ts

marktani:
create new `Prisma` instance for every request

### DIFF
--- a/examples/file-handling-s3/src/index.ts
+++ b/examples/file-handling-s3/src/index.ts
@@ -34,7 +34,7 @@ server.express.post(
   '/upload',
   fileApi({
     s3: s3client,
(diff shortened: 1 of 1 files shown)


examples/file-handling-s3/src/index.ts
```diff
@@ -12,27 +12,30 @@ const s3client = new S3({
...
+  })
 })
 
 server.express.post(
   '/upload',
   fileApi({
     s3: s3client,
-    prisma,
-  }),
+    getPrismaInstance()
```
divyenduz: This needs to be annotated with `prisma:`

marktani:
Thanks @divyenduz @maticzav 🙏

PR #1941, opened 2018-02-20
chore(docs): clarify usage of .env and now

gjungb:
The `now --dotenv` flag by default uses the file `.env`, so I removed the explicit mentioning of this file.
The example regarding "production" was pointing to the wrong `.env` instead of `.env.prod`.


nikolasburk:
Thanks so much 🎉

gjungb:
@nikolasburk Always a pleasure. Keep up the awesome work.

PR #1951, opened 2018-02-21
Update 03-Permissions.md , add missing getUserId(ctx)

muhajirframe:


nikolasburk:
Thanks so much for catchin ghits @muhajirframe 🙌

PR #1953, opened 2018-02-21
Update 05-Subscriptions.md

oiojin831:
without `AND` the example seems to be not working.
Or am I missing something??


nikolasburk:
You're correct, that was indeed an error in the docs. Thanks so much for catchin it @oiojin831 🙌

PR #1960, opened 2018-02-21
fix: remove trailing slash from configured `baseUrl`

akoenig:
This PR proposes to sanitize the `baseUrl` by removing a possible trailing slash.

Reference: https://github.com/graphcool/prisma/issues/1955


timsuchanek:
Thanks for the PR @akoenig !

akoenig:
You‘re welcome :)

PR #1967, opened 2018-02-23
Follow the example .env file name in code

picosam:


nikolasburk:
Great catch @picosam, thanks! 🙏

picosam:
@nikolasburk no worries; but I think it needs to be changed in all doc versions!

PR #1970, opened 2018-02-23
Fix README structure

joar:
It was previously unclear to hasty readers whether or not the list of issues tracking possible future database support was the list of supported databases.


marktani:
Thanks @joar, I adjusted the README to prevent misunderstandings. 8c2a195f446540cf15547e0d36122db6b0106cca

joar:
Thank you @marktani! Hopefully it'll lead to less disappointments for people who do research in quantity rather than quality.

PR #1914, opened 2018-02-15
fix(cli): fix `prisma local stop` message

peterschussheim:
Closes #1898

## Screenshot

iterm2004

PR #1971, opened 2018-02-24
Fix incorrect message when stopping local cluster

hailwood:
It's rather confusing when you tell the cluster to stop and get greeted with `Booting local development cluster`.

This PR simply changes it to `Stopping local development cluster`

### DIFF
--- a/cli/packages/prisma-cli-core/src/commands/local/stop.ts
+++ b/cli/packages/prisma-cli-core/src/commands/local/stop.ts
@@ -16,7 +16,7 @@ export default class Stop extends Command {
   }
   async run() {
     const docker = new Docker(this.out, this.config, this.env, this.flags.name)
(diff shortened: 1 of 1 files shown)


gjungb:
@hailwood @marktani @timsuchanek this seems to be a duplicate of #1914

hailwood:
Ah that it is,

I did a search over the issues for "Booting" before I created this and given that issue doesn't contain the word "Booting" it never came up :)

Take your pick as to whether you'd like "Stopping local ..." or "Shutting down local ..." and if you prefer the "Shutting down local..." variant then close this issue.

timsuchanek:
Thanks for the PR @hailwood !
It is indeed a duplicate of https://github.com/graphcool/prisma/pull/1914, so I'm closing this.

PR #2021, opened 2018-03-05
Give deploy feedback if secrets are changed

do4gr:
- Fixes https://github.com/graphcool/prisma/issues/1433
- still need to check whether console and cli adjustments are necessary


mavilein:
The following things are missing:

- Update the GraphQL Type`MigrationStepType` with the missing type for the new migration step
- the new migration step should contain the actual secrets that should be set
- `ClientDbQueries` must be added to the interface of the deploy connector
- `ClientDbQueriesImpl` must be added to the implementation of the MySQL deploy connector

PR #2027, opened 2018-03-06
Proper Error when adding relations flagged as ambiguous

do4gr:
Fixes https://github.com/graphcool/prisma/issues/1991


server/deploy/src/main/scala/com/prisma/deploy/migration/inference/UnambiguousRelation.scala
```diff
@@ -0,0 +1,15 @@
+package com.prisma.deploy.migration.inference
+
+import com.prisma.deploy.schema.UpdatedRelationAmbigous
+import com.prisma.shared.models.{Relation, Schema}
+
+object UnambiguousRelation {
+
+  def unambiguousRelationThatConnectsModels(schema: Schema, modelAName: String, modelBName: String): Option[Relation] = {
+    val candidates = schema.getRelationsThatConnectModels(modelAName, modelBName)
```
sorenbs: The name indicates that this filter out relations with names, but this actually happens in MigrationstepsInferrer. Can we make this more clear?

PR #2042, opened 2018-03-08
Schema for Self-Relations is not inferred correctly

do4gr:
- the schema inferrer assigned both fields on a self-relation to relationSide A
- the data seems to be correctly entered during mutations, but upon querying it the same side is returned twice
- this changes the relationSide inference two always return two different sides
- should also fix schemas that have been incorrectly inferred during the next change

Fixes https://github.com/graphcool/prisma/issues/1841
Fixes https://github.com/graphcool/prisma/issues/2071


canercandan:
@do4gr is this PR related with #2071? thanks

do4gr:
@canercandan yes, this should fix your issue.

PR #2051, opened 2018-03-09
Fix typo

johannpinson:


nikolasburk:
Grat catches, thanks a lot @johannpinson 🙌

Issue #2057, opened 2018-03-09
Deleting a type is not indicated in the console output


johannpinson:
# Bug Report
**Current behavior**
If I create a new type and deploy, the console indicate it.
But if i delete it, it's not shown

**Reproduction**
- Add a new type and deploy
- Remove it and deploy again
- Re-add the type to a new test
- Just remove a field to see that this change is well displayed in the console

**Expected behavior?**
The console must indicate that the type is deleted, like it's done for a field, cf screenshot

![capture d ecran 2018-03-09 a 18 15 20](https://user-images.githubusercontent.com/2202989/37220317-8960a54c-23c6-11e8-8d45-fe535d432731.png)

johannpinson:
Closed in favor of the pull request

marktani:
Thanks @johannpinson, looks great 🙂 
Let's close this issue when the PR is merged.

johannpinson:
Hey @marktani 

The bug also touch relation after deleting a model.
![capture d ecran 2018-03-13 a 10 39 27](https://user-images.githubusercontent.com/2202989/37333957-ef97e956-26aa-11e8-93d3-eb1c15cc80c4.png)


I will look soon to update the pull request 😉

johannpinson:
@marktani 

I try to update the fragment here:
https://github.com/graphcool/prisma/blob/2f2a943cd2eba668af97f9472490d65f9f33ceec/cli/packages/prisma-cli-engine/src/Client/Client.ts#L81-L83
By adding `leftModel` and `rightModel` as in `CreateRelation`.

But it seems the server can't accept it:
https://github.com/graphcool/prisma/blob/c77e582b5fd6edc62d18bb03d060f5f2d3d2172a/server/servers/deploy/src/main/scala/com/prisma/deploy/schema/types/MigrationStepType.scala#L106-L108

I haven't the good environment to test and debug locally the prisma server, so I prefer let you to assign this issue to the right person 🙂

Thanks,

marktani:
This was resolved.

PR #2058, opened 2018-03-09
fix(cli): update output for DeleteModele action

johannpinson:
Pull request to fix the issue #2057


timsuchanek:
Thanks for the PR @johannpinson !

PR #2059, opened 2018-03-09
Enable changing a field from or to list

do4gr:
The deploy mutation at the moment does not generate the necessary tables / columns when a scalar field switches to/from list.


server/servers/deploy/src/test/scala/com/prisma/deploy/specutils/DeployTestServer.scala
```diff
@@ -114,10 +116,38 @@ case class DeployTestServer()(implicit dependencies: DeployDependencies) extends
...
       Duration.Inf
     )
     println("Request Result: " + result)
     result
   }
+
+  /**
+    * Deploy a schema and return a project
+    */
+  def deploySchema(project: Project, schema: String): Future[Option[Project]] = {
```
mavilein: This is some duplication of what we already have in `DeploySpecBase.setupProject`. Can you add an `addProject` here and call those methods from `DeploySpecBase` please?

server/servers/deploy/src/test/scala/com/prisma/deploy/specutils/DeployTestServer.scala
```diff
@@ -114,10 +116,38 @@ case class DeployTestServer()(implicit dependencies: DeployDependencies) extends
         )
         .recover {
           case error: QueryAnalysisError => error.resolveError
+          case e                         => sys.error(e.toString)
```
mavilein: What is this supposed to do? I think this should just discard the stack trace and we therefore lose information in the exception that is reported in a test.

server/servers/deploy/src/test/scala/com/prisma/deploy/database/schema/mutations/DeployMutationSpec.scala
```diff
@@ -659,6 +656,66 @@ class DeployMutationSpec extends FlatSpec with Matchers with DeploySpecBase {
...
+        |type A {
+        | name: String! @unique
+        | value: Int
+        |}
+      """.stripMargin
+
+    server.query(
+      s"""
+         |mutation {
+         |  deploy(input:{name: "${nameAndStage.name}", stage: "${nameAndStage.stage}", types: ${formatSchema(schema)}}){
```
mavilein: We should rely on the new shiny `server.deploySchema` here as well.

server/servers/deploy/src/test/scala/com/prisma/deploy/specutils/DeployTestServer.scala
do4gr: oversight from debugging, I removed it.

server/servers/deploy/src/test/scala/com/prisma/deploy/specutils/DeployTestServer.scala
do4gr: Done

PR #2068, opened 2018-03-11
Minor Update Doc - Remove `Bearer` prefix on Authorization

Cerberus:
## Update Documentation

- Remove `Bearer` prefix on Authorization (HTTP Header)

## Note

- I've tried to import/export data via `curl` on Prisma v 1.3.5 with `Bearer` prefix and without it and It work both. I've choiced the second one for the docs.

-  Some pages on Graphql Ecosystem (e.g. Graphql Config, Graphql Request) were not updated. I suppose these packages can be used with general graphql server. Not only Prisma.

: )

### DIFF
--- a/docs/1.3/04-Reference/03-Prisma-API/02-Concepts.md
+++ b/docs/1.3/04-Reference/03-Prisma-API/02-Concepts.md
@@ -197,7 +197,7 @@ The GraphQL API of a Prisma service is typically protected by a [service secret]
 The `secret` is used to sign a [JWT](https://jwt.io/) which can then be used in the `Authorization` field of the HTTP header:
 
 ```
-Authorization: Bearer __TOKEN__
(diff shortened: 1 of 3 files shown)


Cerberus:
I have signed the CLA for graphcool/prisma, but CLA-bot does not update(also click recheck). T^T

nikolasburk:
Hey @Cerberus, thanks so much for the PR! 

Since usage of `Bearer ` is a [standard](https://security.stackexchange.com/questions/108662/why-is-bearer-required-before-the-token-in-authorization-header-in-a-http-re), we'll keep including it in our docs. However, to point out that people can also omit it, we're going to add a note that our sever implementation also accepts headers where it's left out. If you have some time, you're also more than welcome to create such a note in the docs, otherwise we're going to add it soon 🙂

Issue #2046, opened 2018-03-08
Incorrect `InvalidNames` being accessed


tbrannam:
When defining relationships, the incorrect error message is displayed in the event of an invalid character being used in the name.

Actual: 
<details><summary>Full Error Response</summary>

```
{
  "data": {
    "deploy": null
  },
  "errors": [
    {
      "locations": [
        {
          "line": 2,
          "column": 9
        }
      ],
      "path": [
        "deploy"
      ],
      "code": 4004,
      "message": "User_UserProfile is not a valid name for a relation. It must start with a letter and may contain up to 30 letters, numbers, underscores and hyphens.",
      "requestId": "cluster:cluster:cjeipnny900ad0152wkx3e04h"
    }
  ],
  "status": 200
}
```
</details>

```
ERROR: User_UserProfile is not a valid name for a relation. It must start with a letter and may contain up to 30 letters, numbers, underscores and hyphens.
```

Expected:

```
Error: The provided name: User_UserProfile is not valid for a relation. It can only have up to 54 characters and must have the shape [A-Z][a-zA-Z0-9]*
```

See: https://github.com/graphcool/prisma/blob/c9ed5f6fefc25bdc2aa55f590135f08d94048bf6/server/deploy/src/main/scala/com/prisma/deploy/schema/Errors.scala#L29

marktani:
This is now fixed and available in `1.4.0` 🙂Thanks for bringing this up, @tbrannam and thanks for the fix @johannpinson 🙌

MaximeScibetta:
When I `prisma deploy`

```

ERROR: The provided name: coursesByDriver is not a valid name for a relation. It can only have up to 54 characters and must have the shape [A-Z][a-zA-Z0-9]*

{
  "data": {
    "deploy": null
  },
  "errors": [
    {
      "locations": [
        {
          "line": 2,
          "column": 9
        }
      ],
      "path": [
        "deploy"
      ],
      "code": 4004,
      "message": "The provided name: coursesByDriver is not a valid name for a relation. It can only have up to 54 characters and must have the shape [A-Z][a-zA-Z0-9]*",
      "requestId": "eu1:cluster:cjgnhmu7gpqe00b182jk6ynw7"
    }
  ],
  "status": 200
}
```

dcasier:
The first character must be uppercase : [A-Z][a-zA-Z0-9]*

PR #2073, opened 2018-03-12
Update error for invalid relation name #2046

johannpinson:
Fix for #2046


do4gr:
Thx for the fix @johannpinson !

PR #2077, opened 2018-03-13
Fix test broken by relation error message fix from johan

do4gr:
fix test broken as a result of https://github.com/graphcool/prisma/pull/2073


johannpinson:
normally it's johann (with a double "n") 😄

do4gr:
I'll try to remember it next time ;-)

PR #2088, opened 2018-03-14
Update cli/README.md for npm package page

johannpinson:
A little update for the README inside the `cli/` folder and which is use for the npm package page.
(like just a ctrl+c/v of the README at the root of the repository 😉)


johannpinson:
I keep the fury badge for `prisma` for "sub"-package to maintain the simplicity from the whole.

In a next version, it can be a good thing to add specific content to each REAME and describe what it the role of every package.

marktani:
Thanks! I agree with your suggestions for the future README as well.

PR #2095, opened 2018-03-15
Do not use SQLActionBuilders in import to reduce memory footprint

do4gr:


server/connectors/api-connector-mysql/src/main/scala/com/prisma/api/connector/mysql/database/SlickExtensions.scala
```diff
@@ -94,6 +94,32 @@ object SlickExtensions {
...
 
+  def escapeUnsafeParamToString(param: Any): String = {
+    def unwrapSome(x: Any): Any = {
+      x match {
+        case Some(x) => x
+        case x       => x
+      }
+    }
+    unwrapSome(param) match {
+      case param: String       => s"'$param'"
```
dpetrick: This doesn't escape anything. Just glancing over the code, this seems like an injection issue here. We need to discuss this in person how we want to utilize proven libraries (e.g. slick helpers or others) to prevent SQL injections.

PR #2136, opened 2018-03-21
I added Google Cloud SQL to the Prisma Cloud docs.

maarteNNNN:
If there's anything wrong, I'm able to edit it. I didn't include images because it's fairly straight-forward.


nikolasburk:
Hey @maarteNNNN, so sorry for the delay! This is really awesome! 🎉 In fact, I think this should rather be two separate tutorials instead of cramping both DB setups into one. The rest of the tutorial can remain the same, only the DB setup part is different.

In the side-bar, I'd also call them specifically according to the DB provider:

- `Prisma Cloud (AWS RDS)`
- `Prisma Cloud (Google Cloud SQL)

Also, even though it might be straightforward, it would be super awesome if you could add some screenshot that depict the flow (not absolutely necessary though, I'm also happy to merge a version without screenshots).

Let me know if you have any questions 🙌

marktani:
I agree that screenshots would be super helpful - a tool like [Monosnap](https://monosnap.com/welcome) makes it easy to create instructional screenshots 🙂

maarteNNNN:
@nikolasburk @marktani Should be all good now ;-).

nikolasburk:
Hey @maarteNNNN, I'm really sorry for the delay! I just merged the PR (there's a slight issue with the alias though as we can't have two documents using the same alias, I'll fix this myself). In any case, thanks so much for contributing this! 🙏 🎉

nikolasburk:
Hey @maarteNNNN, I'm really sorry for the delay! I just merged the PR. I'll add it to the most recent Prisma version in the docs myself!

Issue #2131, opened 2018-03-21
Passing cluster address and secret by cli argument or env var


mzygmunt:
It would be nice if we could pass cluster address and secret by the cli argument or env var. To run the "prisma deploy" command by CI/CD. Then, it would not be necessary to create the file "~/.prisma/ config.yml" in the docker's image. Or by the file "database/prisma.yml".

marktani:
Thanks a lot for your feature request!

This could be done using this command:

```
prisma cluster add --endpoint http://example.com/ --secret mysecret --name mycluster
```

mzygmunt:
@marktani When I try to execute this command, than I get error "Unknown flag --endpoint". Are these parameters in version 1.4.1?

marktani:
My message was a suggestion for introducing these flags. They are not present yet.

mzygmunt:
@marktani I added this functionality as you suggested #2149

marktani:
Thanks for providing the PR, but I'll close this issue due to the deprecation of `prisma cluster add`:

> Thanks for the PR @mzygmunt! We're deprecating the cluster commands, as the management of clusters will change significantly in the upcoming 1.7 release.
> Stay tuned for more information on that tomorrow!

Issue #1948, opened 2018-02-20
Document programmatic deployment of Prisma services


terion-name:
I ask for assistance in programmatic deployment of Prisma services. I'm not posting this at forum because this topic follows #1808 and it is useful for more deep understanding of Prisma for other devs and can lead to documentation enrichment.

I've tried to reverse-engineer `prisma-cli-engine`, but yet had no luck with this.

My goal is to build a cli with node js, that will read multiple Prisma graphql-schemes and will wind up several special Prisma services for further use in a complex system, built on top of them (and also winded up with my cli).

Lets go from a point, where I have a docker mysql service and graphql prisma schemes in place. I need to:
1. create prisma containers
2. join them in to the network with mysql
3. link mysql service to prisma services
4. migrate Prisma service with new or updated schemas

First 3 steps are easy with [docker api](https://github.com/AgustinCB/docker-api) and by reproducing environment as described [here](https://www.prismagraphql.com/docs/reference/clusters/docker-aira9zama5).

So now we have running containers with Prisma and Mysql. How to provision/migrate them? As far as I understand, this can be accomplished with `prisma-cli-engine/Client`, but without documentation and with quite complex code, I can't figure out how.

akoenig:
@terion-name Thanks for you effort to a more programmatic way of deploying Prisma services. I have to ask some questions in order to fully understand what you're trying to achieve: Do you plan to have one MySQL instance and several Prisma service containers that interact with that instance? Why not deploying everything horizontally – one Prisma service + one MySQL instance?

Then you could follow this workflow:

  1. Write a shell script which conducts the MySQL + Prisma service container creation (in your cloud)
  2. Integrate this shell script into your CI/CD pipeline
  3. Write another shell script which adds the cluster to`.prisma/config.yml` and performs a `prisma deploy` afterwards**

With those three steps in place, you can fully automate a Prisma service deployment.

**The configuration file adjustment is a workaround and can be omitted when [#1919](https://github.com/graphcool/prisma/issues/1919) has landed.

terion-name:
Hello, @akoenig 

> Do you plan to have one MySQL instance and several Prisma service containers that interact with that database server instance? If so, why not deploying everything horizontally – one Prisma service + one MySQL instance each?

this is project-specific, long time to explain, that is needed

> Write another shell script which adds the cluster to.prisma/config.yml and performs a prisma deploy afterwards**

prisma config does not allow me to use advanced networking and db linking :(

[...]

Issue #1919, opened 2018-02-16
Non-interactive mode for `cluster add`


akoenig:
# Feature Request

**What feature are you missing?**

It would be great to have a non-interactive mode for `prisma cluster add`. With that in place it would be possible to integrate a `prisma deploy` into a CI/CD process.

**How could this feature look like in detail? Tradeoffs?**

Instead of asking the user in an interactive way, passing the options via CLI arguments could / should also be possible.

marktani:
With the deprecation of `prisma cluster` in `1.7`, I'll close this issue.
We'll make it significantly easier to manage multiple clusters soon.

PR #2149, opened 2018-03-24
Added non-interactive mode for `cluster add`

mzygmunt:
I added possibility to using `cluster add` in non-interactive mode to using command by CI/CD. Please review and merge. 

Ref #1919 #2131 #1948


timsuchanek:
Thanks for the PR @mzygmunt! We're deprecating the `cluster` commands, as the management of clusters will change significantly in the upcoming 1.7 release.
Stay tuned for more information on that tomorrow!

PR #2197, opened 2018-04-06
Introspect existing database and generate sdl

sorenbs:


chaffeqa:
This would be amazing! 😍 

We would love to use prisma, but cannot commit to a non-migration based schema.  This feature would be perfect for us: `let the database schema generate the datamodel.graphql`

PR #2220, opened 2018-04-16
type on line 20

idkjs:


nikolasburk:
Thanks a lot for catching this @idkjs 🙌

PR #2226, opened 2018-04-17
example has incorrect fields on queries

idkjs:
```
type Post
implements Node {
id: ID!
isPublished: Boolean!
title: String!
text: String!
}
```


nikolasburk:
Awesome, thanks a lot for catching this @idkjs 🙌

PR #2281, opened 2018-04-26
rookie friendly edit

idkjs:
On [step 6](https://www.prisma.io/docs/tutorials/cluster-deployment/digital-ocean-(docker-machine)-texoo9aemu), in what shell are we running mkdir? I'm showing docker-machine active to be prisma but running `mkdir` is doing it on my local machine. I created all the files by ssh'ing into the droplet, but i'm feeling like i'm missing something about running these commands inside the docker container. If I run `docker-compose up -d` from inside the container, it says docker-compose is not installed.


nikolasburk:
Thanks for making that more clear @idkjs 🙏

PR #2304, opened 2018-04-29
[doc] correct managementApiSecret placing

kriswep:
This clarifies the documentation for `docker-compose.yml` file.
`managementApiSecret` is a option under environment, as shown in the example `.yml`. 
But it was located under database->default in the description text. This PR corrects that.
Also, the remaining database default options in the text were further indended for clarification.


sorenbs:
Thank you @kriswep! Small mistakes like this can make it really difficult to adopt Prisma, so I really appreciate you taking the time to correct it.

PR #2318, opened 2018-04-30
Update Upsert Docs re Nested Mutation Functionality

medelman17:
Because including nested mutations within an upsert mutation is an expected, core functionality, it should be documented that this is not currently possible.


sorenbs:
Thank you for this contribution @medelman17!

PR #2334, opened 2018-05-03
Add validations and extra warnings to deploy when updating relations

do4gr:
Do not strictly require a relation directive to appear twice anymore. https://github.com/graphcool/prisma/issues/1569 , 
Add a validation when adding an omitted back relation later. (If it's a to one relation the omitted side can only be in one relation in existing data.) 
Add an error if we can not determine which relation operations to take after a schema change. Add a hint to provide relation directives or advance in steps. https://github.com/graphcool/prisma/issues/1836
Added tests for a case described here that was already fixed. https://github.com/graphcool/prisma/issues/1944

Fixes https://github.com/graphcool/prisma/issues/1569 , https://github.com/graphcool/prisma/issues/1836

### DIFF
--- a/server/.buildkite/pipeline.sh
+++ b/server/.buildkite/pipeline.sh
@@ -21,6 +21,9 @@ static=$(printf "    - label: \":mysql: MySql API connector\"
 
     - label: \":scala: subscriptions\"
       command: cd server && ./.buildkite/scripts/test.sh subscriptions mysql
+
+    - label: \":scala: integration-tests-mysql\"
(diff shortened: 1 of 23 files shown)


emmenko:
Thanks guys!! Looking forward to try this out 😍

Do you have an ETA on when this is going to be released?

sorenbs:
@emmenko We have a bi-weekly release cycle for Prisma. The next release is scheduled for Tuesday next week.

PR #2365, opened 2018-05-07
docs: sync docker-compose.yml example in docker/upgrade guide to 1.7

divyenduz:
This PR attempts to bring the `docker-compose.yml` examples in sync with each other on the following pages. 

https://www.prisma.io/docs/reference/prisma-servers-and-dbs/prisma-servers/docker-aira9zama5

https://www.prisma.io/docs/reference/upgrade-guides/upgrading-prisma/upgrade-to-1.7-iquaecuj6b

This sync is more important because when `active: true` is omitted from `docker-compose.yml` file, it leads to a cluster restart loop with `docker logs database_prisma_1` yielding an error regarding `active` being required. 

Also, note that `active` field is to be renamed to `migrations` soon.

### DIFF
--- a/docs/1.7/04-Reference/05-Prisma-Servers-&-DBs/01-Prisma-Servers/02-Docker.md
+++ b/docs/1.7/04-Reference/05-Prisma-Servers-&-DBs/01-Prisma-Servers/02-Docker.md
@@ -57,6 +57,7 @@ services:
             user: root
             password: prisma
   db:
+    container_name: prisma-db
     image: mysql:5.7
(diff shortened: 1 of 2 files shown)


divyenduz:
On the same subject, the currently [auto-generated `docker-compose.yml`](https://github.com/graphcool/prisma/blob/master/cli/packages/prisma-cli-core/src/util.ts#L52-L78) file looks incomplete. We should also bring that in sync with these files. 

Please correct me if I am missing something. 

Thanks 👍

PR #2421, opened 2018-05-14
Small doc fixes

SpaceK33z:
Fixes for "Build a GraphQL server for an existing Prisma service",

- Typo: "yoyur" should be "your"
- "mkdir -p my-yoga-server/src" since it is assumed we already have a directory with `prisma.yml` and `datamodel.graphql` in it, maybe just make it `mkdir -p src` instead?
- In the `.graphqlconfig.yml`, it should be `prisma: prisma/prisma.yml` instead of `prisma: prisma.yml`.
- In the `src/index.js` example, `typeDefs: './schema.graphql',` should be `typeDefs: 'src/schema.graphql',`
- In `src/schema.graphql`, `authorID` should be spelled with a lower case `d`; `authorId`.
- In `src/index.js`, `puslished` should be spelled as `published`
- "Filter posts for "GraphQL Europe" this example query should come before the `deletePost` mutation I think, else you don't have any posts to show.


nikolasburk:
Thanks so much for your help @SpaceK33z 🙌

SpaceK33z:
No problem, I wish I had this tutorial when I started with Prisma 😄. It's very good!

PR #2440, opened 2018-05-15
Update 01-Build-a-GraphQL-Server-with-Prisma.md

veksen:


veksen:
will open another PR to avoid the merge pollution

PR #2443, opened 2018-05-16
Remove --project from post-hook command

kitze:
I think this is just confusing. The command still works without specifying the project name, in this case `--project db`. I just pasted that command including `--project db` and got an empty `generated` folder. Now that I have removed `--project db` everything works.


nikolasburk:
Thanks a lot @kitze 🙌

PR #2463, opened 2018-05-18
Fix broken link

sidoshi:
The link to docker reference points to the localhost instead of the Prisma site.


nikolasburk:
Hey @sidoshi, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #2464, opened 2018-05-18
Docs: Correct error language for adding required field

timwis:
This should help users googling or using CTRL+F on the page.


marktani:
thanks!

PR #2474, opened 2018-05-19
:bug: Add missing license to packages

cookiengineer:
This is very important.

:unicorn: 

Thank you.


timsuchanek:
Thanks for the PR! We're looking into it.

PR #2478, opened 2018-05-20
Write tutorial "Connect to Prisma from the frontend"

JulianMayorga:
This tutorial focuses on content mainly. I could have added more screenshots and maybe an initial GIF that shows how the final app looks like. Let me know if you'd prefer the tutorial to have more visual content and I'd happily add it.

I should have added a button on Quickstart that links to this tutorial, but I could not find the Quickstart file that links to the tutorials (The one that corresponds to https://www.prisma.io/docs/quickstart/). I saw it in previous versions of the docs, but not on 1.8.

I will happily receive any kind of feedback for this tutorial 😄


nikolasburk:
This looks great as well, thanks so much 👏

PR #2495, opened 2018-05-23
Update README.md

veksen:
This seemed to be to old README with the old branding. I just 1-to-1 copied the README from the root to here.


timsuchanek:
Thanks for the PR! Haha sometimes it's hard to keep track of all the READMEs

PR #2505, opened 2018-05-24
Update createDraft Mutation resolver

DSM3DIA:
data object shows content: args.title
should be content: args.content


marktani:
thanks!

PR #2531, opened 2018-05-29
Change AWS Fargate supported regions

Siyfion:
AWS added additional regions to the Fargate supported list on April 26th 2018. This updates the tutorial to allow for these other regions.


nikolasburk:
Awesome, thanks a lot @Siyfion 🎉

PR #2541, opened 2018-05-30
update structure to show schema.graphql in /src

captDaylight:


captDaylight:
Would it make sense to add the generated prisma files to a .gitignore? bad idea in general or outside the scope of this tutorial?

nikolasburk:
Thanks for catching this @captDaylight! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #2553, opened 2018-06-01
docs typo on 01-Build-a-GraphQL-Server-with-Prisma.md

giocodes:
reads

> However, this **would you** that you have now two separate locations where these type definitions live.

should be

> However, this **means** that you would have now two separate locations where these type definitions live.


nikolasburk:
Hey @giocodes, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #2555, opened 2018-06-01
Fixed Docker Compose Link

s3cube:
Docker-Compose Reference URL was earlier pointing to localhost:3000, fixed to point to the correct alias


nikolasburk:
Thanks for catching this @s3cube! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #2556, opened 2018-06-01
fix: codegen, downgrade npm-run in prisma cli

divyenduz:
npm-run 5.x is broken on windows
https://github.com/timoxley/npm-run/issues/19


divyenduz:
Moving this PR to alpha.

PR #2559, opened 2018-06-01
Use user provided schema or default to public.

zigomir:
I had issue introspecting a specified schema and then figured out that my selection wasn't even passed to introspect method.

I removed TODO since CLI always asks for schema, so maybe there should be no fallback to `public` or `schemas[0]` which was `public` in my case.


timsuchanek:
Thanks for the PR!

PR #2563, opened 2018-06-02
An example showing how to use GraphQL Shield

christiannwamba:


marktani:
Thanks a ton for this great example!

I just tried to follow the README and at the first mutation `signup`, I encounter an error:

image

For some reason,

```
query:
query ($_v0_where: RoleWhereInput) {
  roles(where: $_v0_where) {
    id
  }
}
operationName: null
variables:
{
  "_v0_where": {
    "name": "USER"
  }
}
```

is returning 

```
{
  "roles": []
}
```

Do we need to add a special seeding script first, that adds those roles?

examples/permissions-with-shield/README.md
```diff
@@ -0,0 +1,123 @@
...
+    content
+  }
+}
+```
+
+You will get an 'Not Authorized` error. This is because you need to be authorized as a user to read posts. Run the following accordingly:
+
+```graphql
+mutation CreateRole {
+  createRole(name: USER) {
```
marktani: we don't need to create a `USER` role because we already did so in our seeding.

examples/permissions-with-shield/README.md
```diff
@@ -0,0 +1,123 @@
...
+    name
+  }
+}
+
+mutation AssignRole {
+  assignRole(role: ADMIN, assigneeEmail: "test@gmail.com") {
```
marktani: is this section supposed to add `ADMIN` or `USER`? this is not clear

examples/permissions-with-shield/README.md
```diff
@@ -0,0 +1,123 @@
...
+    id
+  }
+}
+```
+
+This will create a USER role and assign the role to the just created user.
```
marktani: clean up according to above decision

examples/permissions-with-shield/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+> `assignRole` mutation is open so you don't lock yourself out of your own system :). This kind of mutation should be restricted to only Admins.
```
marktani: in fact the `assignRole` looks like this at the moment:

>     assignRole: rules.isAdmin,

I cannot run this mutation because of this. I am getting `Not Authorised` error.

marktani:
I added the `ADMIN` role to make the `assignRole` mutation run.

However, when I now run 

```graphql
mutation AssignRole {
  assignRole(role: ADMIN, assigneeEmail: "test@gmail.com") {
    id
  }
}
```

I get this error:
```

```

marktani:
Thanks a lot for this great example @christiannwamba 🎉

PR #2568, opened 2018-06-04
fix: prisma seed via run

divyenduz:
Fixed typo in import statement of `cross-spawn`


divyenduz:
Had derived this branch from master. Will reopen after deriving it from alpha.

Issue #2210, opened 2018-04-11
Json variables are not validated


marktani:
# Bug Report

**Current behavior**

Json is not correctly validated.

**Reproduction**

```graphql
type MyRequiredJson {
  id: ID! @unique
  json: Json!
}
```

replace `__ENDPOINT__`:

```sh
curl '__ENDPOINT__' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: https://eu1.prisma.sh' -H 'json: {}' --data-binary '{"query":"mutation b($json: Json!) {\n  createMyRequiredJson(data: {json: $json}) {\n    id\n  }\n}", "variables": {"json": " "}}' --compressed
```

This sets the `json` value to `" "`, which is invalid.

**Expected behavior?**

An error is thrown, saying that `" "` is not a valid value for `Json`.

marktani:
I put the variables in the HTTP headers instead the variables field. 💡

SHASHANK-PRO-05:
@marktani I wanted to contribute to this issue. Can you inform me about the location where to look into the code?

marktani:
Hey @SHASHANK-PRO-05, I suggest to set up a failing test case as the first step, then you can step through the code to see the entire code path that is executed 🙂 

You can add a new `JsonVariablesSpec` file. Just create a new file and add a new test case similar to [this spec](https://github.com/prismagraphql/prisma/blob/master/server/servers/api/src/test/scala/com/prisma/api/queries/SingleItemQuerySpec.scala).

You can use the `query` method and its `variables` parameter for the Json variable test case. Here is [the function](https://github.com/prismagraphql/prisma/blob/master/server/servers/api/src/test/scala/com/prisma/api/ApiTestServer.scala#L28).

do4gr:
This fix is available in stable now.

PR #2610, opened 2018-06-09
#2210

SHASHANK-PRO-05:
This is a WIP in progress Pull Request for issue #2210 
I tried to replicate the issue and found some interesting findings.

1. I have created a test case [here](https://github.com/SHASHANK-PRO-05/prisma/blob/%232210/server/servers/api/src/test/scala/com/prisma/api/queries/JsonVariablesSpec.scala). Let me know if this valid for this issue.
1. If I pass query like that, [here](https://github.com/prismagraphql/prisma/blob/6d37cdd79bc584e573962defdfa884691bc63c73/server/servers/api/src/main/scala/com/prisma/api/schema/CustomScalarTypes.scala#L49) it will be parsed as JsString.
1. But [here](https://github.com/prismagraphql/prisma/blob/6d37cdd79bc584e573962defdfa884691bc63c73/server/servers/api/src/main/scala/com/prisma/util/coolArgs/GcConverters.scala#L22) there is no JsString case match. Due to that, I get an error: java.util.NoSuchElementException: Bad(InvalidValueForScalarType(" ",Json)).get
1. But if I add a case for JsString I can replicate the issue.

**Questions:**
1) Have I done something wrong in creating the query :grin:, which is looking promising as I am not able to replicate the issue with provided steps? But still, I can get into such scenario where Internal server issue is there "Whoops. Looks like an internal server error. Search your server logs for request ID." I think this needs to be handled.
1) According to the code, JsNumber and JsBoolean is also a scenario, and they also don't have any case match in GCConverters. This also needs to be taken care off I guess?
1) If we need to solve them how should I go about solving this?

**Note:**

The test case works perfectly if I pass valid json for $json variable.


do4gr:
Hey, @SHASHANK-PRO-05  thanks for looking into this. 
Your reproduction looks good to me. I get the same error when passing the query. Let me think a bit where we want the Json validation to happen and then I'll get back to you!

marktani:
To me it looks like your test case tests the _current behaviour_ - but the current behaviour is wrong!

Current behaviour: `"Empty JSON " should "be validated "`
Expected behaviour: `"Empty JSON " should "be be detected as invalid "`

do4gr:
I had a look and we were passing in wrong conversion logic to Sangria (the GraphQL library we are using up front) in coerceUserInput within CustomScalarTypes. 

We were passing in logic that was accepting way to many options and did not have a failure case. I made a PR with a fix. You can have a look at it and update your PR with it and we'll merge yours. https://github.com/prismagraphql/prisma/pull/2613

As you can see in CustomScalarTypes, we now only accept JsObject and JsArray in CoerceUserInput. For all other values we let Sangria throw a JsonCoercionViolation which is the behavior we want since it now errors at the earliest possible stage.

For me it was helpful while debugging to also write test cases that should pass in order to see which codepaths they take. And we of course always want tests for successful cases as well to prevent regressions anyway. So next time that might help you to just write a couple more tests and step through them to get a feeling for the code. 

This was a tricky issue for a first contribution, so thanks for trying!

SHASHANK-PRO-05:
@do4gr @marktani I have merged the pull request #2613 with this pull request.

do4gr:
Thanks for the PR @SHASHANK-PRO-05 , merging this now.

PR #2614, opened 2018-06-11
Clean up of GCValues

do4gr:
- clean up the GCValues since a lot of their usages have been refactored

### DIFF
--- a/server/servers/deploy/src/main/scala/com/prisma/deploy/gc_value/GcConverters.scala
+++ b/server/servers/deploy/src/main/scala/com/prisma/deploy/gc_value/GcConverters.scala
@@ -15,9 +15,21 @@ import scala.util.control.NonFatal
 import scala.util.{Failure, Success}
 
 /**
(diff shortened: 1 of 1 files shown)


server/libs/gc-values/src/main/scala/com/prisma/gc_values/GcValues.scala
```diff
@@ -36,27 +32,28 @@ case class RootGCValue(map: SortedMap[String, GCValue]) extends GCValue {
...
+  def isEmpty: Boolean   = values.isEmpty
+  def size: Int          = values.size
+  def value: Vector[Any] = values.map(_.value)
 }
 
-sealed trait LeafGCValue                    extends GCValue
-object NullGCValue                          extends LeafGCValue
+sealed trait LeafGCValue extends GCValue
+object NullGCValue extends LeafGCValue {
+  def value = None
```
mavilein: Why must this be a `None`? I would have rather expected `null`.

PR #2632, opened 2018-06-14
Make examples more resiliant

geirman:
The latest version of node does not support object spread unless you pass the `--harmony` flag. Since I expect many people will run into this issue, I thought it made sense to add the harmony flag to the command by default. Couldn't hurt, right?

Also, named queries and mutations help prevent the "This anonymous operation must be the only defined operation" error from rearing it's ugly head.

### DIFF
--- a/docs/1.9/03-Tutorials2/02-Build-GraphQL-Servers/01-Development/01-Build-a-GraphQL-Server-with-Prisma.md
+++ b/docs/1.9/03-Tutorials2/02-Build-GraphQL-Servers/01-Development/01-Build-a-GraphQL-Server-with-Prisma.md
@@ -406,7 +406,7 @@ The implementation of your GraphQL server is now done and you can work with it i
 In your terminal, start the server using the following command:
 
 ```sh
-node src/index.js
+node --harmony src/index.js
 ```
 
 </Instruction>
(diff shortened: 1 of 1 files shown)


docs/1.9/03-Tutorials2/02-Build-GraphQL-Servers/01-Development/01-Build-a-GraphQL-Server-with-Prisma.md
```diff
@@ -406,7 +406,7 @@ The implementation of your GraphQL server is now done and you can work with it i
 In your terminal, start the server using the following command:
 
 ```sh
-node src/index.js
+node --harmony src/index.js
```
peterschussheim: I think a better suggestion would be to add a key with `engine` version in `package.json` that specifies the min node version.

https://docs.npmjs.com/files/package.json#engines

geirman: As you wish, but 8.2.1 is the latest (recommended) version per node's site. And it works, but requires the --harmony flag. I can see your point too though. Maybe both since neither one on it's own is a complete solution?

nikolasburk:
Hey @geirman, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now (since the tutorials now live [here](https://www.prisma.io/tutorials)) so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #2669, opened 2018-06-20
Not prompted, given endpoint

captDaylight:
When I run the command:

`graphql create permissions-example --boilerplate node-advanced`

I'm not prompted to choose a public endpoint, but rather given one like so:

```
Your Prisma GraphQL database endpoint is live:

  HTTP:  https://eu1.prisma.sh/public-example-123/permissions-example/dev
  WS:    wss://eu1.prisma.sh/public-example-123/permissions-example/dev
```


nikolasburk:
Thanks a lot for fixing this and sorry for the delay 🙏

PR #2682, opened 2018-06-24
Update schemPath -> schemaPath

drop-paul:


nikolasburk:
Thanks a lot for fixing this and apologies for the delay @drop-paul 🙌

PR #2697, opened 2018-06-27
Remove the cluster and local command

divyenduz:
- Removes some commented code
- Removes related usage of command imports
- Adds a new class CommandRemovedError to handle removed deprecated
commands in future


divyenduz:
Fixes https://github.com/prismagraphql/prisma/issues/2666

divyenduz:
Moving to alpha.

PR #2703, opened 2018-06-28
Update the structure of my-yoga-server

Ferdous-Al-Imran:


nikolasburk:
Thanks a lot for catching this and sorry for the delay @Ferdous-Al-Imran 🙌

PR #2704, opened 2018-06-28
Rephrase a bit.

Ferdous-Al-Imran:


nikolasburk:
Thanks a lot for catching this @Ferdous-Al-Imran 🙌

PR #2726, opened 2018-07-03
Updated localhost to prisma.io for link.

happylinks:
Probably this should be an alias, but I have no idea how the docs get build.
Let me know if I can do it better! I couldn't really find it in the contribution guidelines.


nikolasburk:
Thanks a lot for catching this @happylinks 🎉

PR #2734, opened 2018-07-04
[WIP] Small cleanups of JDBC connector

do4gr:


flieks:
is this for sql server connector? 
Then nice work ;)

do4gr:
This is part of our streamlining of the common parts of the sql connectors. Which will make it easier to roll out more Sql flavors.

PR #2739, opened 2018-07-05
fix: typo bigrations -> migrations

DavidJFelix:
I do like the word "bigrations" though


nikolasburk:
Thanks a lot for the fix @DavidJFelix 🙏

*whispers*: also, bigrations are our secret feature yet to be released 😏

PR #2741, opened 2018-07-06
Fix introspection - relations

divyenduz:
This PR re-implements the introspection feature following an evolving spec based on the expected API [outlined here](https://www.prisma.io/docs/reference/service-configuration/data-model/introspection-aeb6diethe/#generated-data-model) and also based on feedback for pervious versions:-

1. Postgres Introspection Bugs https://github.com/prismagraphql/prisma/issues/2504

Apart from the outlined spec, this PR handles
1. Various scalar transformations.
1. Commenting out invalid types / fields.
1. Ambiguous relation mapping.
1. Nullability of fields 

---- 

Other observation related to this feature:- 

- [ ] https://github.com/prismagraphql/prisma/issues/2605 - We still generate `Int!` type for `bigint` but the backend blows up. 
- [ ] The introspection generates `id Int @unique` but the backend blows up for this field. 
screen shot 2018-07-06 at 1 20 49 pm


divyenduz:
- [ ] Still need to handle self relations! from [chinook schema](https://github.com/lerocha/chinook-database/wiki/Chinook-Schema) an `employee` can have multiple `employees` (under him) and an `employee` can `report to` another `employee`. We need to mark these with `@relation` directive. 
screen shot 2018-07-06 at 1 51 34 pm

divyenduz:
- [ ] For the `largeSchema.test.ts` SQL. The following generated (and one another) type is failed by backend with the following error:-

screen shot 2018-07-06 at 2 17 36 pm

```
      "message": "Syntax error while parsing GraphQL query. Invalid input \"\"staff_list\") {\\n  address: String\\n  city:
 String\\n  country: String\\n  id: Int\\n  name: String\\n  phone: String\\n  sid: Int\\n  zip c\", expected IgnoredNoComm
ent or ValueConst (line 195, column 32):\ntype Staff_list @pgTable(name: \"staff_list\") {\n
```

cli/packages/prisma-db-introspection/src/types/graphql.ts
```diff
@@ -0,0 +1,101 @@
...
+  constructor(name: string, fields: GQLField[], directives: string[], renderCommented: boolean) {
+    this.name = name
+    this.fields = fields
+    this.directives = directives
+    this.renderCommented = renderCommented
+  }
+
+  // Determines if this type is valid GraphQL compatible with Prisma
+  isValid(): boolean {
+    return this.fields.some(f => f.isValid())
```
dpetrick: Shouldn't this test for `!f.isValid`?

divyenduz: If a `GQLType` has any valid fields, it is a valid type. 
Invalid fields are commented out making the rest of the type valid.

divyenduz:
Also, this fixes: https://github.com/prismagraphql/prisma/issues/2781

PR #2746, opened 2018-07-09
fix: introspection input - don't ask for schemaName when no data

divyenduz:
Fixes https://github.com/prismagraphql/prisma/issues/2413


timsuchanek:
It turned out that this case is already handled by https://github.com/divyenduz/prisma/blob/01bda96f627fc0a48a5167256f881c8dd521ae0e/cli/packages/prisma-cli-core/src/utils/EndpointDialog.ts#L382-L386

PR #2780, opened 2018-07-18
feat(graphql-auth): auth example rework

abhiaiyer91:


examples/graphql-authentication/.graphqlconfig.yml
```diff
@@ -1,17 +1,17 @@
...
     extensions:
       endpoints:
-        default: 'http://localhost:4466/graphql-authentication'
+        default: http://localhost:4466/graphql-authentication
       prisma: prisma.yml
       codegen:
-        output: 
-          typings: generated-prisma.ts
+        output:
+          typings: src/generated/generated-prisma.ts
```
nikolasburk: let's also change the name of the generated TS files to just `prisma.ts` instead of `generated-prisma.ts`

PR #2664, opened 2018-06-20
Contribute Readme

SHASHANK-PRO-05:
I have written a document that can help developers in getting started with Prisma Development. Let me know the changes.

PR #2785, opened 2018-07-18
Moving it to different branch as earlier one got stale #2664

SHASHANK-PRO-05:
This is an update on #2664. The previous one is stale. @mavilein


mavilein:
Thanks a lot 🙏 !

PR #2908, opened 2018-08-11
Might want to add an error warning...

Esxiel:
Couldn't find the solution through googling, so I think it would help a lot of people who are trying out AWS for the first time! Sorry to not propose a change, since I'm not sure where to make the change(or if maybe its just me)

It might be just me as a first time AWS service(or ECS to be exact) user, but I had an error along the lines of "Unable to assume service linked role" which was baffling at first. 

It turns out that as per https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using-service-linked-roles.html,
I need the AWSServiceRoleForECS role in my "Roles", which is... normally created automatically when you make a cluster, but it didn't for me, and had to make it manually through the CLI as explain in the link on top.

**Also, a lot of regions are now available!


nikolasburk:
Thanks for catching this @Esxiel! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #2916, opened 2018-08-14
Update warning message

mcmar:
The warning message for Management API Security disabled tells users the old way to do things.
Instead, we should tell users the new way to do things using PRISMA_CONFIG.

The contributing guidelines say that PRs for server should be made against the alpha branch. This is technically a PR for server, but it's just updating warning text. If you're ok, it'd be nice to merge directly to master. Please let me know and I'll change the base fork to master.


mcmar:
Also related, if you'd be so kind: https://github.com/prismagraphql/prisma-templates/pull/18

Issue #2772, opened 2018-07-13
CLI command to generate management token for a given management secret


marktani:
**Is your feature request related to a problem? Please describe.**
It's currently not clear how to generate a valid management API token for the management API of a Prisma Server, even if the management API secret is known.

**Describe the solution you'd like**
A CLI command, for example `prisma management-token`, that can be used to generate a valid management API token.

This command should not require any file to be present in the current folder, and has no assumptions about `~/.prisma/config.yml`.

The command requires an environment variable called `PRISMA_MANAGEMENT_API_SECRET` to be set, or alternatively a secret can be passed using an argument (see below). If a secret is passed via an argument, the argument has precedence over the environment variable.

The command has these arguments:

* an optional argument called `--secret|-s`  to pass a managent API secret. this has precedence over the env var `PRISMA_MANAGEMENT_API_SECRET` and the `-e` parameter.
* an optional argument, `-e, --env-file` to pass in an env file. The only env var that is looked for in this file is the `PRISMA_MANAGEMENT_API_SECRET` variable
* an expiry duration in hours, called `--expiry-duration|-d`
  * a reasonable default expiry duration could be 1 hour
* an optional argument called `--copy|-c` to copy the result to the clipboard instead of printing it to `stdout`.

The output of this command is a JWT with this payload, signed with the management API secret:

```json
{
  "grants": [
    {
      "target": "*/*",
      "action": "*"
    }
  ],
  "iat": 1531498832,
  "exp": 1689286832
}
```

**Describe alternatives you've considered**

Maybe `--expiry-duration` can be dropped in favor of the default value of 1 hour.

**Additional Context**

The expected shape and process to create a management API token should be documented in any case (independently from this command).

jhalborg:
This would be great for devops work! :)

PR #2939, opened 2018-08-17
CLI command to generate management token for a given management secret

w0wka91:
https://github.com/prisma/prisma/issues/2772


w0wka91:
Just added some tests. 
The command generates a token with an expiry duration of 1 hour if no other duration is passed.

szepeviktor:
@w0wka91 @marktani Is this PR already in prisma? https://github.com/prisma/prisma/blob/master/cli/packages/prisma-yml/src/Cluster.ts#L76-L104

PR #2943, opened 2018-08-19
Update Zeit Now + GCP example to use Cloud SQL Proxy

sammccord:
* Modify existing external IP-related documentation with cloud sql proxy instructions
* Add relevant instruction, mainly around creating service accounts and adding secrets.
* Update Dockerfile to use multi-stage builds, and run cloud_sql_proxy in parallel with prisma server.

I was very excited when I saw the initial tutorial pop up, as I was developing a similar approach for GCP projects behind an all-Now service cluster.

The only adjustment I'm suggesting is replacing the external IP connection with one through the official cloudsql proxy. It's better security practice to control access with IAM roles, use SSL, and not manually authorize external networks. Also, it's pretty easy to implement so why not!


nikolasburk:
Hey @sammccord, thank you so much for the PR and my sincere apologies for the delay! Unfortunately, I wasn't able to merge the PR in time and we're now already at docs version 1.17. Would you be able to make the changes to the current docs version 1.17 and 1.8 as well so I can merge them?

frandiox:
@sammccord Thanks a lot for this guide, even though it wasn't merged. Works quite well!

PR #3025, opened 2018-09-05
feat(client): fragment support

divyenduz:
Add support for custom fragment to `prisma-lib`.


divyenduz:
Released to alpha via https://github.com/prisma/prisma/pull/3026

PR #3039, opened 2018-09-07
Documentation on PRISMA_MANAGEMENT_API_SECRET ENV var when Docker

ShunnyBunny:
I had issue where my local Docker will not work until I set my PRISMA_MANAGEMENT_API_SECRET env variable. If this is indeed needed, it is currently missing from the doc every where I looked.


nikolasburk:
Hey @ShunnyBunny, we're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now (since the tutorials now live [here](https://www.prisma.io/tutorials)) so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #3046, opened 2018-09-09
[docs] Fix examples using ApolloClient

brikou:
`import { ApolloClient } from 'apollo-boost'` does not work it should be `import ApolloClient from 'apollo-boost'`.

BTW, I've also import `gql` from apollo-boost (instead of graphql-tag).


nikolasburk:
Hey @brikou, thanks so much for creating this PR! We're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #3049, opened 2018-09-09
fix: typo referencing mysql instead of postgresql

thefliik:
There was a typo referencing MySQL instead of PostgreSQL


nikolasburk:
Hey @thefliik, thanks so much for creating this PR! We're currently cleaning up our PRs. Apologies for not having looked at it in detail earlier, unfortunately it is outdated by now so I'll close it. Please keep opening PRs in the future if you see any issues with our docs 🙏

PR #3050, opened 2018-09-09
a typo

mrkvon:
a single letter typo in docs
 `authentivate` -> `authenticate`


mrkvon:
Ah, what a drag! :(

Some words about granting patents which I don't understand. Therefore can't sign. What's their purpose?

Perhaps somebody else can commit this small typo, then?

PR #3083, opened 2018-09-11
Fix typo in docs

jpJuni0r:


marktani:
The current version is the correct markdown syntax. The link does not work because it's inside `<InfoBox>`

PR #3090, opened 2018-09-12
Steps to migrate from prisma binding

LawJolla:
Here were the steps I followed to move to `Prisma Client` from `Prisma Binding`.

I know nothing about formatting the front matter for the docs, and my markdown is ability is 0.  I'm hoping I've provided the steps where someone that knows markdown can properly format it and make the appropriate front matter changes.

### DIFF
--- a/docs/1.17/prisma-client/migration/Migrate-from-Prisma-Bindings
+++ b/docs/1.17/prisma-client/migration/Migrate-from-Prisma-Bindings
@@ -1,0 +1,78 @@
+import Collapse from "components/Markdown/Collapse"
+import Code from "components/Markdown/Code"
+import Warning from "components/Markdown/Warning"
+
+export const meta = {
+  title: 'Prisma Binding Migration (JavaScript)',
+  position: 20,
+  technology: 'node',
+  technologyOrder: 1,
+  articleGroup: 'Binding Migration',
+}
+
+## Overview
+
+For Prisma 1.17 and greater, `Prisma Client` provides the next evolution of Prisma bindings.  Migrate from < 1.17 Prisma Bindings to Prisma Client with the following steps.
+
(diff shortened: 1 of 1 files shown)


docs/1.17/prisma-client/migration/Migrate-from-Prisma-Bindings
```diff
@@ -0,0 +1,78 @@
...
+
+```js
+const { Prisma } = require("prisma-binding")
+
+...
+
+const db = new Prisma({
+  resolvers,
+  typeDefs: `./generated/prisma.graphql`,
+  ...
```
brikou: You should write `// ...` instead of `...`

brikou: everywhere please :)

schickling:
Thanks a lot for opening this PR @LawJolla. We're currently updating the docs to make it easier to understand when to use Prisma client vs Prisma binding. We'll get back to this PR soon!

LawJolla:
Thanks @schickling !  I see you guys went in a different direction with the API, so I was going to close this.  But I can take a second swing once the API stabilizes.

nikolasburk:
Hey @LawJolla, thanks so much for opening this PR and the thought you've put into it! After an internal discussion we've decided to not merge it. The main reason for that is that we see the client and bindings to be used in distinct scenarios (more info in [this](https://www.prisma.io/forum/t/help-understanding-prisma-clients-value-proposition/4394/17?u=nikolas) forum post) so that there shouldn't actually be a need to "migrate" from one approach to the other. We are working on more "official" resources (other than the linked forum post) that will make the usages of both approaches more clear!
Thanks again for taking the time to create this PR and thinking along, this is very much appreciated! 🙏

PR #3101, opened 2018-09-14
Example code is missing comma

jamesmbowler:
I was going through the example here and was getting an error when trying to run node index.js.


marktani:
I think there might be something wrong with your merge, can you create a new PR that only changes the comma instead of other files? 😄

jamesmbowler:
Sorry, and looks like someone already fixed it ..

marktani:
Ah right, I just merged a similar PR. Thanks so much for your contribution either way 🙏

PR #3112, opened 2018-09-17
WIP: fix(client): post launch feedback

divyenduz:
- Client needs to be bumped and published


client/src/codegen/typescript-definitions.ts
```diff
@@ -0,0 +1,12 @@
+import { TypescriptGenerator } from './typescript-client'
+
+export interface RenderOptions {
```
timsuchanek: Why are you adding special `RenderOptions` here?

PR #3137, opened 2018-09-19
client: sort types before generating CRUD schema

### LINKED ISSUE #3102: `prisma generate` commit or not commit generated files?
**Describe the bug**

Imagine 2 people are working on the same project, personA works on branA and personB on branchB.
PersonA make a minor change datamodel.graphql, run `prisma generate`
PersonB make also a minor change datamodel.graphql, run `prisma generate`

BranchA is merged, then branchB cannot be merge without running `prisma generate` again. On each `prisma generate`, generated code is shuffled and so conflict happens almost at 100%.

So there is 2 options to prevent collisions:

- don't commit generated files
- prevent as much as possible collision

IMO generated files should be commited, as part of code source (event if generated)

**How to prevent collisions**

In order to prevent collision, we need to sort generated types, scalars, enums by their name. As a bonus point having thing sorted would allow to check generated files, when running `git add --patch`, and so would make any regression detected sooner.
Another thing, if we can expect the result, someone can patch generated files for custom need (like adding tslint, eslint annotation, fe).

WDYT?

brikou:
fixes #3102

### DIFF
--- a/cli/packages/prisma-generate-schema/src/index.ts
+++ b/cli/packages/prisma-generate-schema/src/index.ts
@@ -20,7 +20,10 @@ export function parseInternalTypes(model: string) : IGQLType[] {
  */
 export function generateCRUDSchema(model: string) : GraphQLSchema {
   const generators = new Generators()
-  const types = parseInternalTypes(model)
+  const types = parseInternalTypes(model).sort(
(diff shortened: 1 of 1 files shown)


brikou:
@timsuchanek can you give me your feedback ?

PR #3148, opened 2018-09-19
add info on host.docker.internal in documentation

jamesmbowler:
Seems a lot of people are unable to connect to local db from docker. I know there are multiple ways of doing this, but host.docker.internal seems the most common.

### DIFF
--- a/docs/1.17/get-started/01-setting-up-prisma-existing-database-a003.mdx
+++ b/docs/1.17/get-started/01-setting-up-prisma-existing-database-a003.mdx
@@ -34,7 +34,7 @@ Using your existing database with Prisma currently only works when using **Postg
 
 Make sure to have connection details for your database at hand. This includes the following pieces of information:
 
-- **Host**: The host of your Postgres server, e.g. `localhost`.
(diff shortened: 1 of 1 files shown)


nikolasburk:
Thanks so much for opening this PR and apologies for the delay. Would you mind updating the PR with two minor changes:

- add parantheses around both added sentences: "The host of your Postgres server, e.g. `localhost`. (When connecting to a local database, you may need to use `host.docker.internal`.)"
- also add this to the new [1.18](https://github.com/prisma/prisma/blob/master/docs/1.18/get-started/01-setting-up-prisma-existing-database-a003.mdx) version of the docs

nikolasburk:
@jamesmbowler I just added this to the new docs as well. Thanks so much for your input 🙏

lastmjs:
Hey, so `host.docker.internal` apparently only works on Windows and Mac, see here: https://github.com/docker/for-linux/issues/264

I spent a long time trying to figure this out, and it would be great to put that in the documentation somehow. My workaround, and I'm not sure how generally applicable this is, is to set the host to `172.172.17.0.1`.

Issue #3184, opened 2018-09-25
Client generation with a live endpoint


divyenduz:
**Is your feature request related to a problem? Please describe.**
Currently, we use the `prisma-generate-schema` package to generate a Prisma schema based on the datamodel without a live endpoint. While this solution is awesome because we don't need a running service anymore but it is also another abstraction which makes tracking a bunch of issues difficult - because we now have two places to check. 

**Describe the solution you'd like**
We can use a flag to signal to `prisma generate` command to not use the `prisma-generate-schema` and use a live endpoint or pre-generated schema instead. Example: `prisma generate --use-endpoint`.

**Additional context**
We need to test that how this feature lines up with known API differences https://github.com/prisma/prisma/issues/3185

Because things like `executeRaw` are artificially hidden in schema generator and might be exposed when we bypass it.

schickling:
I really like this idea. How about the following syntax/behavior:

- Default (based on datamodel): `prisma generate`
- Using `endpoint` in `prisma.yml`: `prisma generate --endpoint`
- Using custom endpoint: `prisma generate --endpoint http://some-server:4466`

divyenduz:
The syntax `prisma generate --endpoint <url>` is moved to this feature request:- https://github.com/prisma/prisma/issues/3206

divyenduz:
This fix is available in the stable version - `1.18.1` - `npm install -g prisma` 🎉

Issue #3185, opened 2018-09-25
Client generated API and GraphQL API differences


divyenduz:
**Is your feature request related to a problem? Please describe.**
The generated client and the underlying Prisma GraphQL API are not a 1-1 mapping. The client attempts to simplify some cases enabling users to write less code for a few operations. 

**Typescript Client**

1. Create mutations

GraphQL
```
mutation ($data: TodoCreateInput!) {
  createTodo(data: $data) {
    id
    title
    createdAt
  }
}
```

Client
```
await prisma.createTodo({
  title,
})
```

Notice that client simplifies it by removing the `data` wrapper. 

2. Delete mutations

GraphQL
```
mutation ($where: TodoWhereUniqueInput!) {
  deleteTodo(where: $where) {
    id
    title
    createdAt
  }
}
```

Client
```
await prisma.deleteTodo({
  title,
})
```

Notice that client simplifies it by removing the `where` wrapper.

3. Queries with only one argument which is `where`

Similarly to the above two examples, Prisma client simplifies the API by removing the singular variable wrapper. 

4. `$raw` and `executeRaw`

Prisma GraphQL API exposes a special mutation called `executeRaw`. The client does not expose the `executeRaw` mutation and instead exposes the functionality on `client.$raw` namespace. 

**Go Client**

1. Go Client `nil` type safety does not match 1-1 with GraphQL schema. Here is the relevant discussion: https://github.com/prisma/prisma/issues/3288

**Describe the solution you'd like**
While the new shorter API is convenient to use and does the same thing with less code. It deviating from the underlying GraphQL might cause some confusion. Particularly, in JS only codebases when usually there would be no autocompletion etc. Also, when we fall back to raw GraphQL for a task, the deviation from the GraphQL API might lead to confusion. 

This issue tracks the discussion around this topic and we want community feedback on how beneficial this deviation is and if we need to make amends to it.

simonjoom:
hello 
I tried prisma-client one thing about this 'writelesscode' 
Yes indeed was a good point but for me gave me some headache more than supposed to resolve;
Before with prisma-binding we used forwardTo who is great because args are consistent between the layers.

I explain why here in resolver:

just about the where thing:

if i query user from dashboard
args= { where: { id: userId } }

but i have to add some line of code 
to transform it in 
args={id: userId} 
 to query with prisma client.

with forward in resolver just 
User: forwardTo('db'),
is convenient because args are the same. It's just avoid lot of headache i think.

[...]

PR #3203, opened 2018-09-27
feat(schema-generation): bypass js generator via flag

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3184

1. - [x] Add `--endpoint` flag
1. - [ ] Add `--endpoint <url>` flag (tracked via https://github.com/prisma/prisma/issues/3206)
1. - [x] Handle `executeRaw` mutation field (also consider - https://github.com/prisma/prisma/issues/3185)


divyenduz:
This is ready to merge. Point `2` can be handled separately as a feature, it is a big change to add that with the current CLI code because:-

1. No good API to dynamically change/create prisma definition based on an endpoint. 
1. No good API to support both `string`, `boolean` type for one flag endpoint.

schickling:
Sounds good @divyenduz. Can you create a separate issue for this?

Issue #3149, opened 2018-09-19
Only one Workspace is selectable during ‘prisma init’


wes-cutting:
TL;DR : I would like to initialize a Prisma service in a second Workspace and am unable to select the region from CLI ‘prisma init’ process.

As I run through a new ‘prisma init’ I select the option ‘Demo server’ under the category ‘Or deploy to an existing Prisma server’. Before this process, I had logged in through the CLI and can see that it is linked to the right account as it shows the first Workspace I created. However, that is the only Workspace that is available from the CLI. So when it asks for me to select a region, I only see the first Workspace.

Is this intended behavior? If it isn’t, how do I reveal my other regions/workspaces so I can initialize a new service outside of my first environment cluster?

If more clarification is needed please let me know.

timsuchanek:
![image](https://user-images.githubusercontent.com/1094804/46136858-c991a780-c248-11e8-8b79-bf9c41c2f665.png)

It's already enough information. When selecting the demo server, only the demo servers for one workspace are shown. But actually all workspaces should be available in the selection.

divyenduz:
This fix is available in the stable version - 1.18.1 - `npm install -g prisma` 🎉

PR #3204, opened 2018-09-27
fix(init): show demo servers from all workspaces

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3149

I removed the sort logic from CLI. But looks like the servers are already sorted and grouped by workspaces which was the desired state (also, tried with hardcoded opposite latency). 

screen shot 2018-09-27 at 4 30 41 pm

### DIFF
--- a/cli/packages/prisma-cli-core/src/utils/EndpointDialog.ts
+++ b/cli/packages/prisma-cli-core/src/utils/EndpointDialog.ts
@@ -642,23 +642,22 @@ export class EndpointDialog {
   }
 
   private async askForDemoCluster(): Promise<Cluster> {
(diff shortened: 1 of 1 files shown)


marktani:
Does this also show other servers hosted on all workspaces?

divyenduz:
This PR specifically solves the issue when the user chooses "Demo server". Production grade clusters are already working in stable prisma CLI. 

screen shot 2018-09-27 at 4 51 06 pm

For e.g. above the "Demo Server" option, we have two servers coming from separate workspaces.

marktani:
Ah that's great, I misunderstood the original issue. 💯

PR #3209, opened 2018-09-28
Update management-api-foe1.mdx

lidorcg:
- fix typo duye -> due


nikolasburk:
Thanks for catching this @lidorcg! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3214, opened 2018-09-29
Fix forwardTo function

nekocode:
[`forwardTo`](https://github.com/prisma/prisma/blob/master/client/src/utils/index.ts#L76) function will access a wrong legacy location in the prisma client object. These modifications fix it.

### DIFF
--- a/client/src/utils/index.ts
+++ b/client/src/utils/index.ts
@@ -80,18 +80,18 @@ export function forwardTo(bindingName: string) {
     context: CONTEXT,
     info: GraphQLResolveInfo,
   ) => {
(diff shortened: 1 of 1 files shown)


divyenduz:
@nekocode Thanks for raising this PR, I really appreciate the work you have put in, however, exposing `forwardTo` from client was not intended. This PR removes `forwardTo` from Prisma Client: https://github.com/prisma/prisma/pull/3264

The reason for that is `forwardTo` uses schema delegation and we recommend using `prisma-binding` for that. 

Please follow this for a detailed answer: https://www.prisma.io/forum/t/help-understanding-prisma-clients-value-proposition/4394/17?u=divyendu_singh

Thanks!

Issue #3205, opened 2018-09-27
Prisma login should work in environment without a browser, such as codesandbox.io


sorenbs:
**Describe the bug**
When using the Prisma CLI in the new https://codesandbox.io container environment, signing in fails with error code 3. This happens because the login flow requires the CLI to open a browser window, which is not available in the container environment.

To work around the issue, you can manually copy the login url, but because the CLI returns an error, it is not longer listening for the successful login.

**To Reproduce**
1. Get access to the new https://codesandbox.io container environment (currently invite only)
2. Install Prisma
3. Use the CLI to login

**Expected behavior**
Instead of returning an error, the CLI should tell the user to copy/paste the url, and keep listening for successful login.

divyenduz:
This is resolved in `1.25.0-beta.5` (install via `npm install -g prisma@beta`) via this PR https://github.com/prisma/prisma/pull/3842

Here is a sample sandbox that can be used to test it: https://codesandbox.io/s/91qj216zny

divyenduz:
This might still yield this message incorrectly on some operating systems, see https://github.com/sindresorhus/opn/issues/86

However, this won't stop them from logging in.

ecker00:
Thank you, that worked for me! 🎉 With the `@beta` installed. Very useful when running in a headless vm.

divyenduz:
This is available in the latest version `1.25.0`, `npm install -g prisma`. I am closing this issue. Please let me know if the issue still persists via opening a new issue.

PR #3232, opened 2018-10-02
Prisma login should work in environment without a browser, such as codesandbox.io

w0wka91:
Await opn instead of catching and rethrowing the error.
Related to this issue https://github.com/prisma/prisma/issues/3205


divyenduz:
@w0wka91 : Thanks for the PR 🙌 I really appreciate your time and effort into fixing this. 

One minor change, I updated the message to be more friendlier, can you please bump the snapshot again 👍

divyenduz:
Thanks for raising this PR 🙌

However, I am closing this in favor of this: https://github.com/prisma/prisma/pull/3842

I could not make this one work on an actual browser-less environment. I think that the `await` never returns in this case.

PR #3249, opened 2018-10-03
Move @types/jest to devDependencies

marcqualie:
Having these types pulled in makes prisma unusable with something like mocha.

All other prisma packages have jest in dev rather than main dependencies.

Example errors:

```
Variable 'beforeEach' must be of type 'Lifecycle', but here has type 'HookFunction'
Variable 'it' must be of type 'It', but here has type 'TestFunction'
```

### DIFF
--- a/cli/packages/prisma-generate-schema/package.json
+++ b/cli/packages/prisma-generate-schema/package.json
@@ -14,12 +14,12 @@
   "author": "Emanuel Jöbstl",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@types/jest": "^23.3.1",
     "jest": "^23.5.0",
     "ts-jest": "^23.1.4",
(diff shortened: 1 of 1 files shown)


divyenduz:
@marcqualie : Thanks for the PR. Nice catch.

marcqualie:
@divyenduz do you have an estimate on when this might make it into a release?

Still a blocker for us to get past 16.x at all due to the conflicts with `chai` when compiling typescript

divyenduz:
@marcqualie : This is already available in beta. `npm install -g prisma@beta` or install beta version of the specific package. 

This will hit stable on 23rd October according to our release policy. 
https://www.prisma.io/blog/improving-prismas-release-process-yaey8deiwaex/

Issue #3248, opened 2018-10-03
Client: Does not support models with name "Type`


ryanquinn3:
**Describe the bug**
The new prisma client maintains an [internal data structure that has the name 'types'](https://github.com/prisma/prisma/blob/master/client/src/Client.ts#L40). It appears as if each of the root query types from prisma is also added to the client as properties. This can cause naming collisions.

If one were to have a model in their `datamodel.graphql` with the name `Type` then prismas root query would have a field on it named `types`

```graphql
type Type @model {
  id: ID! @unique
}
```
generates a schema with:
```graphql

type Query {
   types(where: TypeWhereInput, orderBy: TypeOrderByInput, skip: Int, after: String, before: String, first: Int, last: Int): [Type]!
}
```

The naming collision occurs [here](https://github.com/prisma/prisma/blob/master/client/src/Client.ts#L311) and the original `client.types` is overridden. 

**Expected behavior**
Prisma client would not impose any naming restrictions on model names. 

**Screenshots**
You can see here that `client.types` is the function that fetches the `Type` model.
![screen shot 2018-10-03 at 2 48 50 pm](https://user-images.githubusercontent.com/5115565/46442011-27c4fb80-c71d-11e8-804a-93ec7949a96a.png)

**Versions (please complete the following information):**
 - OS: `OS X High Sierra` 
 - `prisma` CLI: 1.18
 - Prisma Server: 1.17
 - `prisma-client-lib`: 1.17.1

**Additional context**
If the prisma team is okay with it, I'm happy to put up a PR that adds a `_` prefix to the members the of `Client` class.

timsuchanek:
Thanks for reporting @ryanquinn3. That is indeed the case! It would be awesome if you could create that PR. Let me know if you have any questions!

Nedomas:
Also ran into this. As a temporary hack you can call `prisma.buildMethods()` just right before your shadowed `prisma.types(...` call. Hoping to have @ryanquinn3's PR merged soon.

timsuchanek:
The PR has been merged and released in stable :)

PR #3252, opened 2018-10-04
Underscore client internals as to not clash with downstream schemas

ryanquinn3:
Fixes: https://github.com/prisma/prisma/issues/3248

- Prefix internal `Client` fields with `_` as to prevent naming collisions from the callers schema


timsuchanek:
Thanks for the PR @ryanquinn3 and sorry for the delay! It will land in stable today.

Issue #3213, opened 2018-09-29
`prisma init` should not require `docker-compose` when using the demo server


sorenbs:
When running `prisma init` and selecting the demo server, Prisma CLI will complain about `docker-compose` not being present on computers without Docker Compose. This does not make sense, as Docker COmpose is not required when using the demo server.

<img width="831" alt="screen shot 2018-09-29 at 17 30 25" src="https://user-images.githubusercontent.com/281337/46247554-6309b680-c40d-11e8-9340-100395744d71.png">

PR #3231, opened 2018-10-02
`prisma init` should not require `docker-compose` when using the demo server

w0wka91:
Docker compose installation will only be checked when set up a new Prisma server or deploy to an existing server. https://github.com/prisma/prisma/issues/3213

PR #3291, opened 2018-10-10
don't write docker-compose.yml for demo server, only ask for docker-compose if needed

marktani:
see #3213


marktani:
Oops just saw there is already a PR https://github.com/prisma/prisma/pull/3231 by @w0wka91. It takes a slightly different approach.

PR #3295, opened 2018-10-10
Corrected express instantiation in Docs

heyMP:
Fixes
```
/Users/michael/github/prisma-test/index.js:2
const app = express()
            ^

ReferenceError: express is not defined
```


nikolasburk:
Thanks for catching this @heyMP! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3307, opened 2018-10-14
Fixed example for Basic filters for lists

eperedo:
In the example the code is:

```js
const linksBeforeChristmas = await prisma
  .links({
    where: {
      createdAt_lt: "2017-12-24" 
    }
  })
```

But the description says: **Fetch links created before December 24, 2018:**


marktani:
thanks!

PR #3308, opened 2018-10-14
Adding missing "s" in graphql config file

eperedo:
I was having a hard time generating the graphql schema from the prisma service. I am new in graphql and did not now [that according the specification the property name is "projects"](https://github.com/prisma/graphql-config/blob/master/specification.md), but in the example it says  "project".

Hope it helps newbies like me 😄


marktani:
Thanks!

PR #3327, opened 2018-10-16
fix(go): add auth

divyenduz:
Add auth support for Prisma service protected with a secret.

### DIFF
--- a/client/src/codegen/go-client.ts
+++ b/client/src/codegen/go-client.ts
@@ -650,6 +650,24 @@ export class GoGenerator extends Generator {
     }
   }
 
+  printSecret(options: RenderOptions): string | null {
+    if (!options.secret) {
+      return `""`
+    } else {
+      if (options.secret!.startsWith('${process.env')) {
+        // Find a better way to generate Go env construct
+        const envVariable = `${options.secret!
+          .replace('${process.env[', '')
+          .replace(']}', '')}`
+          .replace("'", '')
+          .replace("'", '')
+        return `os.Getenv("${envVariable}")`
+      } else {
+        return `\"${options.secret.replace("'", '').replace("'", '')}\"`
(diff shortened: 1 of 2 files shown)


client/src/codegen/go-client.ts
```diff
@@ -705,16 +723,23 @@ type Client struct {
 }
 
 type Options struct {
-	Endpoint string
```
dominikh: Inconsistent indentation, mixing tabs and spaces. Before the change, all the code in the `static` string was indented with tabs.

client/src/codegen/go-client.ts
```diff
@@ -705,16 +723,23 @@ type Client struct {
...
+  Secret   string
 }
 
 func New(options *Options, opts ...graphql.ClientOption) *Client {
-	endpoint := DefaultEndpoint
+  endpoint := DefaultEndpoint
+  secret   := Secret
 	if options != nil {
-		endpoint = options.Endpoint
+    if options.Endpoint != "" {
```
dominikh: Has what I discussed with @schickling changed? We had agreed that if the user provided a non-nil `options`, that all values would have to be provided, and not have defaults.

client/src/codegen/go-client.ts
```diff
@@ -650,6 +650,24 @@ export class GoGenerator extends Generator {
     }
   }
 
+  printSecret(options: RenderOptions): string | null {
+    if (!options.secret) {
+      return null
```
dominikh: Wouldn't this branch result in us printing `var Secret = null`? That wouldn't be correct.

client/src/codegen/go-client.ts
divyenduz: Indeed, you are correct, I will make amends.

PR #3337, opened 2018-10-18
Added FAQ Question: Does Prisma Cloud come with a built-in Database?

DATADEER:
I added the FAQ Question: "Does Prisma Cloud come with a built-in Database?" to the Technical section of the FAQ.
I think this is a common misconception for beginners, who tend to use the demo server and expect things to work the same on their own servers.


nikolasburk:
Thanks for adding this @DATADEER! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3338, opened 2018-10-18
fix 3328: change directory before automatically generating client

### LINKED ISSUE #3328: Generate workflow broken for prisma init in a new folder
**Describe the bug**
Currently, the automatic `prisma generate` workflow for `prisma init` is broken for `prisma init <folder-name>`.
screen shot 2018-10-16 at 6 24 28 pm

**To Reproduce**
1. run `prisma init new-folder`
2. see error

**Expected behavior**
> we should `cd` into the folder and change this to the programmatic implementation that runs the command directly without spawn

**Versions (please complete the following information):**
 - `prisma` CLI: `prisma/1.19.0-beta.8 (darwin-x64) node-v10.4.0`

marktani:
fixes #3328 

still using `spawnSync` as I don't know what was meant by "the programmatic implementation that runs the command directly without spawn" from the issue

### DIFF
--- a/cli/packages/prisma-cli-core/src/commands/init/index.ts
+++ b/cli/packages/prisma-cli-core/src/commands/init/index.ts
@@ -231,6 +231,11 @@ ${chalk.bold('Next steps:')}
 ${steps.map((step, index) => `  ${index + 1}. ${step}`).join('\n')}`)
 
     if (results.generator && results.generator !== 'no-generation') {
+      try {
+        process.chdir(this.config.definitionDir)
+      } catch (err) {
+        this.out.log(chalk.red(err))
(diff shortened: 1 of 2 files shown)


divyenduz:
@marktani:  Ah! I guess that it programmatically changes the directory and generates the client but after the execution the directory is still the same. Is this your understanding as well?

marktani:
`process.chdir` changes the working directory of _the Node process_. So if after the automatic `generate` there would be another command that relies on the direction, it might fail.

However, `process.chdir` does not change the working directory of _the shell_. Therefore point 1 is still valid.

All in all this should not be a problem..

PR #3339, opened 2018-10-18
Added comma in README so that the example code is valid.

nicu-chiciuc:
The code in the README wasn't valid and that's not a good thing.


marktani:
Thanks! This was already fixed here: https://github.com/prisma/prisma/pull/3341

PR #3378, opened 2018-10-26
fix typo

matthewpick:


nikolasburk:
Thanks for catching this @matthewpick! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3383, opened 2018-10-26
Cascading Delete Loop Bug

do4gr:
* make onDelete directive sides stable on self relations
* start lifting limitations on onDelete directive


server/servers/deploy/src/main/scala/com/prisma/deploy/migration/inference/SchemaInferrer.scala
```diff
@@ -173,8 +173,17 @@ case class SchemaInferrerImpl(
       val model1OnDelete: OnDelete.Value = relationField.cascade
       val model2OnDelete: OnDelete.Value = relatedField.map(_.cascade).getOrElse(OnDelete.SetNull)
 
-      val (modelA, modelAOnDelete, modelB, modelBOnDelete) =
-        if (model1 < model2) (model1, model1OnDelete, model2, model2OnDelete) else (model2, model2OnDelete, model1, model1OnDelete)
+//      val (modelA, modelAOnDelete, modelB, modelBOnDelete) =
+//        if (model1 < model2) (model1, model1OnDelete, model2, model2OnDelete) else (model2, model2OnDelete, model1, model1OnDelete)
```
mavilein: Let's remove the commented code.

server/connectors/api-connector-jdbc/src/main/scala/com/prisma/api/connector/jdbc/impl/DeleteNodeInterpreters.scala
```diff
@@ -94,38 +94,45 @@ trait CascadingDeleteSharedStuff {
...
-        idGroup <- groupedIds
-      } yield {
-        recurse(mutationBuilder, field, idGroup, visitedModels :+ model)
-      }
-      _ <- DBIO.seq(childActions: _*)
+      _ <- if (ids.isEmpty) {
+            DBIO.successful(())
+          } else {
+            //children
+            val cascadingChildren = model.cascadingRelationFields.filter(_ != parentField.relatedField)
```
mavilein: Can we suffix this variable with `Fields`? From the name it was not clear to me that this about fields. The name made me think these are nodes.

server/connectors/api-connector-jdbc/src/main/scala/com/prisma/api/connector/jdbc/impl/DeleteNodeInterpreters.scala
```diff
@@ -94,38 +94,45 @@ trait CascadingDeleteSharedStuff {
...
+            val childActions = for {
+              field   <- cascadingChildren
+              idGroup <- groupedIds
+            } yield {
+              recurse(mutationBuilder, field, idGroup)
+            }
+            //other parent
+            val cascadingParent = model.cascadingRelationFields.filter(_ == parentField.relatedField)
```
mavilein: This can always just find one field, right? If this is true, a call to `find` or `headOption` after the filter would make that more obvious.

mavilein: Also the name of the variable is a bit misleading maybe. This is a field on the child model and points to the parent. Right now it reads as if it is a parent field actually.

server/connectors/api-connector-jdbc/src/main/scala/com/prisma/api/connector/jdbc/impl/DeleteNodeInterpreters.scala
```diff
@@ -94,38 +94,45 @@ trait CascadingDeleteSharedStuff {
...
+            val parentActions = for {
+              field           <- cascadingParent
+              idGroup         <- groupedIds
+              idGroupFiltered = idGroup.filter(x => !seenIds.contains(x))
+            } yield {
+              recurse(mutationBuilder, field, idGroupFiltered, parentIds)
```
mavilein: The variables `field` and `idGroupFiltered` both refer to the child model in this case. At the same time we are using `parentIds` with refers to the parent model. The id and filtering can just work if the parent and child model are the same. Did i understand that correctly.

PR #3395, opened 2018-10-30
update example code

willm78:
Renamed the Prisma client instance from `db` to `prisma` to maintain consistency throughout the docs.


nikolasburk:
Thanks for updating this @willm78! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3397, opened 2018-10-31
Correcting relation directive example typo

adamgoose:


nikolasburk:
Hey @adamgoose, thanks so much for the PR. In this case the `@relation` directive requires two times the same name because it's a self-relation 🙏

adamgoose:
Gah, you're right. Ignore me. lol sorry

PR #3400, opened 2018-10-31
Minor fixes to docs

willm78:
replace `schemaPath` with `typeDefs`


Uniixx:
There is an typo in your GraphQLServer const.

nikolasburk:
Thanks for again for the PR @willm78! 😄 While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

Issue #3429, opened 2018-11-06
Prisma client - generate enum types


maxprilutskiy:
**Is your feature request related to a problem? Please describe.**
Prisma client should take advantage of TypeScript's `enum`s instead of just using `type`

**Describe the solution you'd like**
I would like to make codegen module generate TypeScript `enum`s from GraphQL's enums, instead of `type`s.

Advantages:
1. More obvious;
2. Utilizes TypeScript language features;
3. Improves compatibility with other libraries related to code generation.

prilutskiy:
I am going to submit a PR with this change

schickling:
Thanks a lot for opening this issue and suggesting this change. This is a little bit tricky since the `enum` TS syntax and the currently supported string literal syntax are not compatible (this is by design - [see this Github issue](https://github.com/Microsoft/TypeScript/issues/17690)).

Your proposed change would result in a breaking change which would in most cases require a more explicit syntax. Consider the following example based on the giving `datamodel.prisma` snippet:

Datamodel:

```graphql
enum Color { Red, Blue }
```

Resulting TS client types using literal string syntax (current implementation):

```ts
type Color = 'Red' | 'Blue'

// allows for the following
const c1 = 'Red' // ok
const c2 = 'Green' // compile error
```

Resulting TS client types using enum syntax (your suggestion):

```ts
enum Color {
  Red: 'Red',
  Blue: 'Blue'
}

// allows for the following
const c1 = 'Red' // compile error
const c2 = 'Green' // compile error
const c3 = Color.Red // ok
```

---

There is however a workaround described [in the comments of the mentioned issue](https://github.com/Microsoft/TypeScript/issues/17690#issuecomment-350863594) which would allow both the `'Red'` and the `Color.Red` syntax. See the following:

```ts
export type Color = 'Red' | 'Blue'
export const Color: {
  Red: 'Red',
  Blue: 'Blue',
} = {
  Red: 'Red',
  Blue: 'Blue',
}

// allows for the following
const c1 = 'Red' // ok
const c2 = 'Green' // compile error
const c3 = Color.Red // ok
```

While my suggested workaround wouldn't use the canonical `enum` syntax you've described, it would still allow for both referencing syntaxes. Does this solve your requirements?

VRspace4:
Any progress on this? The difference in the way Prisma generates the enum (`string unions`) versus my GraphQL client's way (`typescript enum`) causes conflicts in the types.

chrisui:
[...]

PR #3438, opened 2018-11-07
Enum generation fix in Typescript codegen

prilutskiy:
Updated codegen to generate typescript enums (**not union types**) from graphql enums.

Before:
```
  GraphQLEnumType: (type: GraphQLEnumType): string => {
      if (type.name === 'PrismaDatabase') {
        return ``
      }
      return `${this.renderDescription(type.description!)}export type ${
        type.name
      } = ${type
        .getValues()
        .map(e => `  '${e.name}'`)
        .join(' |\n')}`
    },
```
```
export type UserRole = "ADMIN" | "USER";
```

After:
```
  GraphQLEnumType: (type: GraphQLEnumType): string => {
      if (type.name === 'PrismaDatabase') {
        return ``
      }
      return `${this.renderDescription(type.description!)}export enum ${
        type.name
      } {\n ${type
        .getValues()
        .map(e => `${e.name} = '${e.value}'`)
        .join(',\n')}\n}`
    },
```
```
export enum UserRole {
  ADMIN = "ADMIN",
  USER = "USER"
}
```

Given that the following `enum` is declared in GraphQL:
```
enum UserRole {
  ADMIN
  USER
}
```

### DIFF
--- a/client/src/codegen/typescript-client.ts
+++ b/client/src/codegen/typescript-client.ts
@@ -143,12 +143,12 @@ export type ${type.name}_Output = string`
       if (type.name === 'PrismaDatabase') {
         return ``
       }
-      return `${this.renderDescription(type.description!)}export type ${
+      return `${this.renderDescription(type.description!)}export enum ${
         type.name
-      } = ${type
+      } {\n ${type
         .getValues()
-        .map(e => `  '${e.name}'`)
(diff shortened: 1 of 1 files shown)


schickling:
On hold while still discussing in #3429

PR #3443, opened 2018-11-08
fix inflector

timsuchanek:


cli/packages/prisma-generate-schema/src/util/inflector/english.ts
```diff
@@ -245,20 +427,29 @@ export class English extends TwoFormInflector {
    * @return plural form of the word
    */
   public getPlural(word: string) {
-    return super.getPlural(word);
+    const plural = super.getPlural(word)
```
timsuchanek: This is the core change of the PR.

PR #3544, opened 2018-11-18
'differen' to 'different'

recss:


nikolasburk:
Thanks a lot for catching this @recss 🙌

PR #3548, opened 2018-11-19
Grammatical Error

LeighCiechanowski:
needs to contains at least two properties

to

needs to contain at least two properties


nikolasburk:
Thanks for catching this @LeighCiechanowski! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3554, opened 2018-11-19
minor edits

oreillyross:
The text should refer to Boolean and not to Float, I suspect it was a copy-paste omission to change from Float to Boolean.


nikolasburk:
Thanks for catching this @oreillyross! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3560, opened 2018-11-20
Allow passing absolute path to seed.import

denkristoffer:


denkristoffer:
I'll redo this on `alpha`.

PR #3595, opened 2018-11-26
Fix link to Prisma Client docs

zingerj:
Looks like there is an extra `.` in the link that was breaking it. This should fix the link to the Client docs.


nikolasburk:
Thanks a lot for catching this @zingerj 🙏

Issue #3613, opened 2018-11-27
Flow errors in generated code


HriBB:
**Describe the bug**
Flow complains about unused imports and about `process.env['PRISMA_ENDPOINT']` usage.

**To Reproduce**
1. Use `flow-client` generator
2. Run `prisma deploy`

**Expected behavior**
Flow does not give any errors.

**Screenshots**
![flow_error1](https://user-images.githubusercontent.com/4412324/49104393-27a92e00-f27e-11e8-932b-fd17629ce3c5.png)
![flow_error2](https://user-images.githubusercontent.com/4412324/49104394-27a92e00-f27e-11e8-8bfa-e9534f254085.png)

**Versions (please complete the following information):**
 - `OS` Ubuntu 18.04
 - `flow` **0.86.0**
 - `prisma` 1.21.1
 - `graphql-cli` 2.16.4
 - `prisma-binding` 2.2.8

**Additional context**

I can make a PR to fix this. I tried to use a custom generator but I see it's not so simple.

I had quite a few problems with the flow generated code for `prisma-binding`, here's my override:
```js
// remove unused imports

// export PrismaInterface
export interface PrismaInterface {...}

// remove BindingConstructor

// simply export Prisma otherwise it complains about BindingConstructor
export const Prisma = makePrismaBindingClass({ typeDefs })
```

Would be cool if we could somehow hook into generation process and manipulate the output. You could add overrides in prisma.yml or .graphqlconfig.yml.

divyenduz:
This is fixed in beta version `1.25.0-beta.4`. Please give it a go via `npm install -g prisma@beta` and let me know if that works for you. Thanks.

divyenduz:
This is available in the latest version `1.25.0`, `npm install -g prisma`. I am closing this issue. Please let me know if the issue still persists via opening a new issue.

PR #3658, opened 2018-12-05
fix(flow): flow and env

divyenduz:
Partially fixes https://github.com/prisma/prisma/issues/3613

I am not sure about the approach though and I would request a review from someone who is strong with flow type system. The issue happens because of the following line (variable name can vary)
`${process.env.PRISMA_ENDPOINT}`

Flow cannot cast it back to string because this might be null or undefined. 

Solution 1: 
`${process.env.PRISMA_ENDPOINT || ""}`

For type inferring to work properly. Although this is a hacky solution.

Solution 2: 
Ignore this case by using a special flow comment. 

Review needed. 

P.S. Flow has limited support for tagged template literals: https://github.com/facebook/flow/issues/6996, https://github.com/facebook/flow/issues/4514, https://github.com/facebook/flow/issues/2616, https://github.com/facebook/flow/issues/2814


divyenduz:
Closing in favor of https://github.com/prisma/prisma/pull/3836

PR #3661, opened 2018-12-06
relation name should be Following?

HaveF:


nikolasburk:
Thanks for you PR @HaveF 🙏 In this case, the `@relation` directive does require the same name twice because it's a self-relation.

PR #3690, opened 2018-12-11
Indexes for PostgreSQL

pimeys:
For PostgreSQL this should work quite nicely. Does creation and deletion, altering the index name is almost no-op and uses the `ALTER INDEX` syntax.

## Findings for MySQL:

* Only version 5.7 supports `RENAME INDEX` syntax
* If using fields of type TEXT (+ all subtypes) or BLOG (+ subtypes), the `CREATE INDEX` should look like:

```
mysql> desc bar;
+-------+------------+------+-----+---------+-------+
| Field | Type       | Null | Key | Default | Extra |
+-------+------------+------+-----+---------+-------+
| lol   | text       | YES  |     | NULL    |       |
| foo   | int(11)    | YES  |     | NULL    |       |
+-------+------------+------+-----+---------+-------+
3 rows in set (0.00 sec)
```

```SQL
CREATE INDEX foobar ON bar (lol, foo);
```

This will not work and MySQL needs to know how many characters from the text field is included in the index. It is possible to use `FULLTEXT` index for the text fields, but this index cannot have columns of type which is not text.

One can also do:

```SQL
CREATE INDEX foobar ON bar (lol(N), foo);
```

where N is the number of characters taken to the index. This has some implications... First one is that it'll be a non-cool and a bit complex code to add the size to the fields that are known to be text. One can look into the model fields and detect the type from there, or go to the database to see the type.

Another part is the collisions, especially when we don't have a unique index here. This means whatever text is longer than the N given, and whenever we have another row with the same N characters that differs after the N, we get wrong results back from the database.

## MongoDB

Doesn't support index renaming, so the only way to do it is to drop and create the index. The same needs to be done with MySQL 5.6 and older, and it should be decided that if renaming an index with PostgreSQL is cheap, but with MySQL and MongoDB might cause a day of downtime, what is the right thing to do.


server/servers/deploy/src/main/scala/com/prisma/deploy/migration/validation/DataModelValidatorImpl.scala
```diff
@@ -174,9 +175,10 @@ case class DataModelValidatorImpl(
 
   def validateTypeDirectives(): Seq[DeployError] = {
     for {
-      objectType <- doc.objectTypes
-      directive  <- objectType.directives
-      validator  <- TypeDirective.all
+      fieldAndType <- allFieldAndTypes
```
mavilein: this looks like a copy paste error

server/servers/deploy/src/main/scala/com/prisma/deploy/migration/validation/DeployResults.scala
```diff
@@ -18,6 +18,10 @@ object DeployWarnings {
 }
 
 object DeployErrors {
+  def indexesAreNotSupported(objectType: ObjectTypeDefinition): DeployError = {
+    error(objectType, "Indexes are not supported for the database")
```
mavilein: We should use the term connector instead of database.

PR #3711, opened 2018-12-15
Update link: #graphql-directives → #sdl-directives

chadoh:
This page has a broken internal link. Here's a fix.


nikolasburk:
Thanks for catching this @chadoh and apologies for the delay! While we're now on a newer version of Prisma and the content is outdated, it's still a valuable fix for developers that are using an older version  🙏

PR #3759, opened 2018-12-21
[WIP] use MySQL instead of MariaDB driver

mavilein:


derekperkins:
Is there anything preventing this change from moving forward?

nitins64:
Hi @derekperkins,

what is the current status of supporting vitess from prisma?

I interested in trying out making schema changes to vitess sharded via prisma. is this supported?

nitins64:
Hi @derekperkins ,

any update on support fro vitess using prisma?

derekperkins:
@nitins64 We've stopped trying Prisma, so I don't know for sure. Vitess just understands the MySQL protocol, so if this PR gets merged to use a more modern connector, it should work fine.

nitins64:
It doesn't look like this PR is going to fix the issue. 

It seems Prisma issue following sql to understand the database schema of the mysql db:

>>
  public async listSchemas(): Promise<string[]> {
    const res = await this.query(
      `SELECT 
         schema_name
       FROM 
         information_schema.schemata
       WHERE schema_name NOT LIKE 'information_schema';`,
    )
>>

Problem is that VtGate hides information_schema so prisma get following error:
>>>
 _    Could not connect to database. ER_UNKNOWN_ERROR: vtgate: http://nitin-sharma.lan:15001/: no keyspace in database
 _    name specified. Supported database name format (items in <> are optional): keyspace<:shard><@type> or
 _    keyspace<[range]><@type>
>>>

mavilein:
This will be fixed by Prisma 2 where we will use the plain MySQL driver.

Issue #3725, opened 2018-12-18
Reuse the internal datamodel representation in the introspection


timsuchanek:
In order to support directives like `@db`, `@id` `@createdAt` and `@updatedAt` in introspection, we need to take the current mappings that are defined in the datamodel into account, as not all of the information can be inferred by the pure data and data structure.

The desired way of handling this is the following: When performing the introspection, we will look into the local datamodel file to look for directives that are defined on types or fields that we just introspected. If anything relevant is being found in that local datamodel, we apply that to the introspection result. This way, users can control the introspection result in the future, when they, for example, add a field through a rails migration.

Using the local datamodel as the basis to perform the introspection can have one disadvantage, though: It could get out of sync with the real database. To circumvent this issue, we want to introduce a `prisma introspect --overwrite` command, which would overwrite the local datamodel and is specified in https://github.com/prisma/prisma/issues/3729.

Directives that affect renaming:
- `@db` - this can either rename a field or a type
- `@id` - the MongoDB connector always maps this to `_id`, even if the field which has the `@id` directive applied may be called `foo`

An example schema we want to support:
```graphql
type User @db(name: "users") {
  foo: ID! @id
  createdAt: DateTime! @createdAt
}
```

The handling of the `id` directive is specified in detail in https://github.com/prisma/prisma/issues/3733

ejoebstl:
I've been thinking a lot about the workflow for this feature lately.

Would it make sense to reuse all directives, comments etc in the existing model?
Or in a more extreme case, just add new fields or types to the model, when re-introspecting?

Issue #3702, opened 2018-12-13
Naming of generated types from MongoDB introspection 


nikolasburk:
The convention to name Prisma models is:
- use singular version 
- start with uppercase letters

MongoDB conventions to name collections are:
- use plural version
- start with lowercase letter

To account for this mismatch, `prisma introspect` for MongoDB should generate a name that accords to the Prisma convention and use the `@db` directive to map to the underlying collection. Or the naming convention to use singular, uppercase versions for models should be lifted in Prisma.

timsuchanek:
Applying `camelize(singularize(collectionName))` for example from the `inflector` library should be sufficient.

PR #3776, opened 2018-12-28
Isomorphic Introspection/Type and Field name Normalization

ejoebstl:
This PR implements *isomorphic introspection* (merging an introspection result with an existing data model) and type name normalization for new fields and types, as discussed in #3725 and #3702.

Changes: 
* The `IntrospectionResult` class has two new methods `renderToNormalizedDatamodelString` and `renderToDatamodelString`. Both accept an existing datamodel as optional parameter. 
* Types and fields are matched according to their name, `@db(name: ...)` and `@_id` directives.
* If a field is matched, the name from the existing data model is kept and `@db`, `@updatedAt`, `@createdAt@` directives are copied.
* For new types, or existing types without any `@db` directive, names are normalized according to prisma conventions. This is also recursively done for embedded types. If the name already follows prisma conventions, nothing is done.  
* Comments, formatting and other directives are **not kept**. This is hard to implement consistently when introspection is run multiple times. 

With these changes, it is possible to do **additive** changes in the database, and simply merge them with an existing datamodel when re-introspecting. **Deleting** fields or types will cause the types to vanish from the datamodel as well. **Re-naming** only works when a `@db` directive is placed before introspecting or the type or field is renamed manually in the datamodel. Otherwise, the type is removed and re-added. I would suggest to always write a backup file when introspecting with an existing datamodel.

For an example how this feature works, please have a look at this [unit test](https://github.com/prisma/prisma/blob/tmp-isomorphic-introspection/cli/packages/prisma-db-introspection/src/__tests__/normalization/existingSchema.ts).


schickling:
@nikolasburk let's make sure to document this properly as well 🙂

PR #3784, opened 2018-12-28
Fixed typo in 'subscribe'.

friek:
The document refers to 'subsribe' instead of 'subsribe' in several
places. This commit fixes that.


nikolasburk:
Thanks a lot for catching this @friek 🙌

friek:
You're very welcome :)

PR #3792, opened 2019-01-03
fix(client types): handling of the Node interface

iquabius:
I started setting up typescript code generation for my resolvers and I noticed that `Prisma.node()` was returning an incompatible `Node` type. So I've checked my Prisma client:

```js
// myapp/src/generated/prisma-client/index.ts
// ...
export interface Node {}

export interface Prisma {
  // ...
  node: (args: { id: ID_Output }) => Node;
}
```

Shouldn't `Node` have an `id` property? What's the reasoning behind the empty interface? It seems to be this way since @timsuchanek [introduced `TypescriptGenerator.ts`](https://github.com/prisma/prisma/commit/ed15ba091557d06be8a655d97547d31523270f67#diff-c3c51ea2b11e04f3643366b68facae67).

I noticed there was a `NodeNode` interface which wasn't used anywhere:

```js
// ...
export interface NodeNode {
  id: ID_Output;
}
```

After making some changes to *prisma-client-lib* to fix the `Node` interface, I got this error from running `compilation.test.js`:

```
prisma/cli/packages/prisma-client-lib/dist/codegen/artifacts/generated_ts.ts (507,18):
  Interface 'PostPromise' incorrectly extends interface 'Node'.
  Types of property 'id' are incompatible.
    Type '() => Promise<string>' is not assignable to type 'string'.
```

It seems `ModelPromise` interfaces shouldn't be extending `Node` directly, only `Promise<Model>`.


cli/packages/prisma-client-lib/src/codegen/typescript-client.ts
```diff
@@ -548,7 +546,7 @@ export const prisma = new Prisma()`
     }
 
     return this.renderInterfaceWrapper(
-      `${type.name}${node ? '' : ''}`,
```
iquabius: Someone [forgot this here](https://github.com/prisma/prisma/commit/630ac8930f1348566a1dc4b4b277dabaf42bc4a6#diff-d1ac2c5edb46cae3a12f4ed0eef31d66).

cli/packages/prisma-client-lib/src/codegen/typescript-client.ts
```diff
@@ -771,7 +771,7 @@ ${fieldDefinition}
         ? `export type ${typeName} = AtLeastOne<{
         ${fieldDefinition.replace('?:', ':')}
       }>`
-        : `export interface ${typeName}${typeName === 'Node' ? 'Node' : ''}${
```
iquabius: `${typeName === 'Node' ? 'Node' : ''}` [was added here](https://github.com/prisma/prisma/commit/5de11f305c0a77d6ff9daf028aba0a68ddd87d58#diff-d1ac2c5edb46cae3a12f4ed0eef31d66), and maybe should have been removed [in this commit](https://github.com/prisma/prisma/commit/7602bfef7062eb7bfc3ec13cae9227571b57168f).

Issue #3631, opened 2018-11-30
Prisma Flow client should generate types instead of interfaces where possible


nikolasburk:
There's currently a weird issue with the Prisma Flow client that can be observed here: https://github.com/prisma/prisma-examples/blob/master/flow/script/src/script.js#L23

It is currently required to explicitly cast a part of the input object to `createPost` to `UserWhereUniqueInput` in order for the Flow typechecker to succeed. This gets solved if you replace the `UserWhereUniqueInput` type definition from being an interface to a type.

morajabi:
Yes please!

divyenduz:
This is fixed in beta version `1.25.0-beta.4`. Please give it a go via `npm install -g prisma@beta` and let me know if that works for you. Thanks.

PR #3817, opened 2019-01-07
fix(flow-client): export type for input types

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3631

The original issue is caused by input types, hence this PR only changes
input types from interfaces to types catering for wherever possible. It also creates a separate codepath for flow. 

This minimal change fixes the [script
example](https://github.com/prisma/prisma-examples/tree/master/flow/script).

More thought needs to be given for converting other interfaces to types
as some of them extend other interfaces which will considerably change
the syntax/implementation in types.


divyenduz:
Closing in favor of https://github.com/prisma/prisma/pull/3833

PR #3882, opened 2019-01-16
Fix broken URL in docs/cli#seed

FWirtz:
The URL is still pointing to the old repo.
Correcting that.

PS Is this the right `base` or should it also target `alpha`?


pantharshit00:
@FWirtz 

Thanks for catching this 👍

Issue #3309, opened 2018-10-15
Prisma Client: Support selection of non-scalar fields


timsuchanek:
If you have this schema:

```graphql
type Query {
  user: User!
}

type User {
  house: House!
}

type House {
  id: String!
}
```

and do this client call:

`client.user()` - what should happen?
Right now an error is thrown, that the query contains an empty selection set.
This is the case, because we only expand on scalar fields for now.
However, with types like the `User` type as in our example, that don't have scalar fields, this approach breaks.

An opinionated, non-trivial solution is needed here.

Here are possible solutions for the problem:

1. Expand the first single relation, if that is not possible, expand the first list relation
2. Expand every field

Let's discuss in this issue, which behavior makes more sense and potentially other solutions.

One thing is clear: We at least need better error handling for this case.

marktani:
Just a side not: Currently, this behaviour breaks `xsConnection` queries.

If you run

```
ctx.prisma.usersConnection()
```

```
Field 'usersConnection' of type 'UserConnection' must have a sub selection.
```

schickling:
I like @marktani's suggestion to query for `__typename`. I'm currently leaning towards only querying it if no other scalar fields are available. Are there any reasons why it should be queried in all cases?

vctormb:
> My suggestion: always query `__typename` automatically for every selection set. This way, there is no error thrown. Don't query any relations automatically.
> 
> If the user wants to query a relation, they should be able to easily do so however. I think this would be `client.user().house()`.

Could you please provide an example?

timsuchanek:
We just discussed the concrete implementation and it should look like this:
Only for types that don't have scalar fields, we'll add the `__typename` in the query.
The actual return type won't change - it's an empty object.

divyenduz:
This PR https://github.com/prisma/prisma/pull/3892 implements it for top level types. Resulting in the following behavior:

`prisma.users()` yields the following query now

```graphql
query ($where: UserWhereInput) {
  users(where: $where) {
    __typename
  }
}
```

However, as a result of this opinionated API https://github.com/prisma/prisma/issues/3629 the generated connection API for a type with no scalars breaks currently, generating the following query:-

```graphql
{
  usersConnection {
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
    edges {
      node
      cursor
    }
  }
}
```

[...]

PR #3899, opened 2019-01-21
fix(client): non-scalar sub selection for connections [BREAKING]

divyenduz:
* fix(client): non-scalar sub selection for connections - lib implementation
* fix(client): non-scalar sub selection for connections - types implementation

Fixes https://github.com/prisma/prisma/issues/3309

~~Note that - the existing types snapshot test cases do not cover this types change.~~ This is done now.


cli/packages/prisma-client-lib/src/Client.ts
```diff
@@ -423,6 +424,14 @@ export class Client {
 
     const type = this.getDeepType(field.type)
 
+    if (isRelayConnection) {
+      let relayConnectionHasScalars = false
+      relayConnectionHasScalars = connectionNodeHasScalars({ type })
```
timsuchanek: Please turn this into `const relayConnectionHasScalars = connectionNodeHasScalars({ type })`.

cli/packages/prisma-client-lib/src/utils/connectionNodeHasScalars.ts
```diff
@@ -0,0 +1,49 @@
...
+        fieldType instanceof GraphQLScalarType ||
+        fieldType instanceof GraphQLEnumType
+    )
+}
+
+export function connectionNodeHasScalars({ type }) {
+    const edgesField = Object.entries(type.getFields())
+        .filter(([, subField]: any) => {
+            return subField.name === 'edges'
+        })
```
timsuchanek: Please use `Object.values` here, then we can save the `.map` call https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values

cli/packages/prisma-client-lib/src/utils/connectionNodeHasScalars.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (edgesField.length === 0) {
+        return false
+    }
+
+    const edgesFieldType = getDeepType(edgesField[0].type)
+    const nodeField = Object.entries(edgesFieldType.getFields())
+        .filter(([, subField]: any) => {
+            return subField.name === 'node'
+        })
+        .map(([, subField]: any) => {
```
timsuchanek: Here the same, please use `Object.values`

cli/packages/prisma-client-lib/src/utils/connectionNodeHasScalars.ts
```diff
@@ -0,0 +1,49 @@
...
+            return subField
+        })
+    if (nodeField.length === 0) {
+        return false
+    }
+    const nodeFieldType = getDeepType(nodeField[0].type)
+    const nodeFieldScalars = Object.entries(nodeFieldType.getFields())
+        .filter(([, subField]: any) => {
+            return isScalar(subField)
+        })
```
timsuchanek: `Object.values`

divyenduz:
Thanks. I made the required changes :)

Issue #3907, opened 2019-01-23
Cannot convert undefined or null to object


steida:
When I run `const foo = await db.web({ id: 'cjr8g5yut00fl0778cr8yxzu7' });` and web does not exists, it will throw an exception:

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Client.extractPayload (/Users/steida/dev/este/node_modules/prisma-client-lib/dist/Client.js:246:24)
    at Client.<anonymous> (/Users/steida/dev/este/node_modules/prisma-client-lib/dist/Client.js:115:52)
    at step (/Users/steida/dev/este/node_modules/prisma-client-lib/dist/Client.js:47:23)
    at Object.next (/Users/steida/dev/este/node_modules/prisma-client-lib/dist/Client.js:28:53)
    at fulfilled (/Users/steida/dev/este/node_modules/prisma-client-lib/dist/Client.js:19:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

steida:
If I call exists before, it works. Is it the right behavior?

pantharshit00:
I am able to reproduce this issue:

![image](https://user-images.githubusercontent.com/22195362/51610031-aa2e3100-1f41-11e9-80ec-6f7e9d57a7c5.png)


Here is my reproduction cc @divyenduz https://github.com/pantharshit00/prisma-issue-3907

divyenduz:
This is fixed on the latest stable `1.25.2`, please install it via `npm install -g prisma`

PR #3916, opened 2019-01-23
fix(client): regression because non-scalar resolution

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3907


divyenduz:
Closing in favor of https://github.com/prisma/prisma/pull/3922

PR #3922, opened 2019-01-24
 fix(client): handle null from server

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3907


cli/packages/prisma-client-lib/src/Client.ts
```diff
@@ -193,6 +193,7 @@ export class Client {
     if (
       !selectionFromFragment &&
       Array.isArray(pointer) &&
+      Boolean(pointer) &&
```
timsuchanek: Does something speak against moving `Boolean(pointer)` up here?

divyenduz: Done! :)

PR #3957, opened 2019-01-30
Subscriptions and Selected Fields

do4gr:
* limit the use of SelectedFields.all in Mongo
* use selected fields from context when fetching subscription results

### DIFF
--- a/server/connectors/api-connector/src/main/scala/com/prisma/api/connector/DataResolverTypes.scala
+++ b/server/connectors/api-connector/src/main/scala/com/prisma/api/connector/DataResolverTypes.scala
@@ -56,8 +56,8 @@ object QueryArguments {
 object SelectedFields {
   val empty                                                             = SelectedFields(Set.empty)
   def forRelationField(rf: RelationField)                               = SelectedFields(Set(SelectedRelationField.empty(rf)))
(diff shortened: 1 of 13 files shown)


server/connectors/api-connector/src/main/scala/com/prisma/api/connector/DataResolverTypes.scala
```diff
@@ -56,8 +56,8 @@ object QueryArguments {
 object SelectedFields {
   val empty                                                             = SelectedFields(Set.empty)
   def forRelationField(rf: RelationField)                               = SelectedFields(Set(SelectedRelationField.empty(rf)))
-  def all(model: Model)                                                 = SelectedFields((model.scalarFields.map(SelectedScalarField) ++ model.relationFields.map(SelectedRelationField.empty)).toSet)
   def byFieldAndNodeAddress(field: RelationField, address: NodeAddress) = address.path.selectedFields(field)
+  def all(model: Model)                                                 = SelectedFields((model.scalarFields.map(SelectedScalarField) ++ model.relationFields.map(SelectedRelationField.empty)).toSet)
```
do4gr: rename this to allScalars, since it should only be used in JDBC

Issue #3951, opened 2019-01-29
Go Client: WhereInput array values are not pointers when generating go client


kevinwmiller:
**Describe the bug**
When using the prisma generated graphql schema and go client with graph-gophers, I receive the error message
```The error in question is "panic: field "Where": field "IDIn": []string is not a pointer```

Here is my datamodel.prisma
```graphql
type User {
	id: ID! @unique
	email: String! @unique
	password: String!
	displayName: String!
	createdAt: DateTime!
	updatedAt: DateTime!
	needsToResetPassword: Boolean @default(value: false)
	spaces: [Space!]
	courses: [Course!]
}

type Space {
	id: ID! @unique
	name: String! @unique
	description: String!
	inviteOnly: Boolean @default(value: false)
	users: [User!]
}
```
The go client contains the following for SpaceWhereInput:

```go
type SpaceWhereInput struct {
	ID                       *string           `json:"id,omitempty"`
	IDNot                    *string           `json:"id_not,omitempty"`
	IDIn                     []string          `json:"id_in,omitempty"`
	IDNotIn                  []string          `json:"id_not_in,omitempty"`
    ...
	UsersNone                *UserWhereInput   `json:"users_none,omitempty"`
	And                      []SpaceWhereInput `json:"AND,omitempty"`
	Or                       []SpaceWhereInput `json:"OR,omitempty"`
	Not                      []SpaceWhereInput `json:"NOT,omitempty"`
}
```
The generated graphql files has following User definition
```graphql
type User {
  id: ID!
  email: String!
  password: String!
  displayName: String!
  createdAt: DateTime!
  updatedAt: DateTime!
  needsToResetPassword: Boolean
  spaces(where: SpaceWhereInput, orderBy: SpaceOrderByInput, skip: Int, after: String, before: String, first: Int, last: Int): [Space!]
  courses(where: CourseWhereInput, orderBy: CourseOrderByInput, skip: Int, after: String, before: String, first: Int, last: Int): [Course!]
}
```
The error appears when implementing a resolver for the Users field, specifically in the Where input of the resolver

I was able to resolve the issue by changing the go client code to the following
```go
type SpaceWhereInput struct {
	ID                       *string           `json:"id,omitempty"`
	IDNot                    *string           `json:"id_not,omitempty"`
	IDIn                     *[]string          `json:"id_in,omitempty"`
	IDNotIn                  *[]string          `json:"id_not_in,omitempty"`
    ...
[...]

PR #3985, opened 2019-02-03
fix(go-client): Optional lists will now be generated as pointers

kevinwmiller:
Fix for Issue #3951


kevinwmiller:
Ah okay. Didn't realize they were were similar to c pointers. Sounds like it's a bug in the other library I'm using then. Thanks for the information!

Issue #3976, opened 2019-02-01
Add "do not edit" warning for top of prisma-schema.ts in the generated folder


arnabkd:
**Is your feature request related to a problem? Please describe.**
I think it is a bit confusing that only one of the generated files in the typescript prisma-client have a `DO NOT EDIT.` warning on them.  

**Describe the solution you'd like**
One of the below:
-  Have the same warning header that `index.ts` has
- Maybe have a `README.md` in the generated folder that contains the warning

**Describe alternatives you've considered**
This is not a huge problem as common sense dictates that one does not edit generated files, so no alternatives are really necessary. 

**Additional context**

divyenduz:
This is available in the latest `beta` - `1.27.0-beta.12`, please install it via `npm install -g prisma@beta`.

divyenduz:
This is available in the latest `stable` - `1.28.0`, please install it via `npm install -g prisma`. I am closing this issue, please open a new issue, if necessary.

PR #3992, opened 2019-02-05
fix(client): add do not edit message on top of ts type defs

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3976


cli/packages/prisma-client-lib/src/codegen/generators/__tests__/typescript-client.test.ts
```diff
@@ -18,3 +18,12 @@ test.only('typescript generator', t => {
   const result = generator.render()
   t.snapshot(result)
 })
+test.only('typescript generator definitions', t => {
```
timsuchanek: Please remove the `.only` :)

divyenduz: Done! :)

PR #3993, opened 2019-02-05
[BREAKING] Finish SQLite

do4gr:
* this disables the ability to deploy renames of the shape ( A->B, B->X) in one go, these will now produce a deploy error and ask to be split into two steps
* if any change on a model happens we now always generate an updateModel step instead of only a RenameModel step if the name changes.


server/connectors/api-connector-jdbc/src/main/scala/com/prisma/api/connector/jdbc/impl/JdbcDatabaseMutactionExecutor.scala
```diff
@@ -18,9 +19,34 @@ case class JdbcDatabaseMutactionExecutor(
     extends DatabaseMutactionExecutor {
   import slickDatabase.profile.api._
 
+  private def runAttached[T](project: Project, query: DBIO[T]) = {
```
mavilein: Can you move that method further down? We should aim at being able to read from top to bottom.

server/connectors/api-connector-sqlite/src/main/scala/com/prisma/api/connector/sqlite/SQLiteDatabasesFactory.scala
```diff
@@ -7,7 +7,6 @@ import slick.jdbc.SQLiteProfile
 import slick.jdbc.SQLiteProfile.api._
 
 object SQLiteDatabasesFactory {
-  private lazy val dbDriver = new org.sqlite.JDBC
```
mavilein: All other database factories hand in the JDBC driver as an argument instead of hardcoding it. I don't know why that is exactly. I think @dpetrick introduced that structure. Can you talk to him to understand why?

server/connectors/deploy-connector-jdbc/src/main/scala/com/prisma/deploy/connector/jdbc/database/ColumnMutactionInterpreters.scala
```diff
@@ -150,54 +110,60 @@ case class UpdateColumnInterpreter(builder: JdbcDeployDatabaseMutationBuilder) e
...
-      case (true, true, true)  => Vector(removeUniqueConstraint, updateColumn, addUniqueConstraint)
-      case (true, _, false)    => Vector(removeUniqueConstraint, updateColumn)
-      case (true, false, true) => Vector(updateColumn)
-      case (false, _, false)   => Vector(updateColumn)
-      case (false, _, true)    => Vector(updateColumn, addUniqueConstraint)
+    def addUniqueConstraint = builder.addUniqueConstraint(mutaction.project, after)
+
+    def updateColumnActions = (before.isUnique, typeChanges, requiredChanges, after.isUnique) match {
+      // type changes, after unique
+      case (true, true, true, true)   => Vector(updateColumn, addUniqueConstraint)
```
mavilein: The very explicit cases did not help me understand what the logic was. But the comment did. After reading the comment a single pattern `case(_, true, _, true)` is more descriptive.

server/servers/deploy/src/main/scala/com/prisma/deploy/migration/inference/MigrationStepsInferrer.scala
```diff
@@ -72,13 +72,10 @@ case class MigrationStepsInferrerImpl(previousSchema: Schema, nextSchema: Schema
       nextModel         <- nextSchema.models.toVector
       previousModelName = renames.getPreviousModelName(nextModel.name)
       previousModel     <- previousSchema.getModelByName(previousModelName)
-      if nextModel.name != previousModel.name || nextModel.isEmbedded != previousModel.isEmbedded
-    } yield UpdateModel(name = previousModelName, newName = nextModel.name)
+      if nextModel != previousModel
```
mavilein: I think we should not compare the full models here but only the template.

Issue #3977, opened 2019-02-02
Please remove red text color for non-errors in "post deploy" section of "prisma deploy" output.


zxol:
Red text in the console should be reserved for (serious) errors.
Unnecessary heartbeat elevation for your users :)

Mercurial:
agreed ahaha!

divyenduz:
This is available in the latest `beta` - `1.27.0-beta.12`, please install it via `npm install -g prisma@beta`.

divyenduz:
This is available in the latest `stable` - `1.28.0`, please install it via `npm install -g prisma`. I am closing this issue, please open a new issue, if necessary.

garmjs:
I'm on prisma/1.28.2 and the red color on the success message is still here, it's better if we change with green

![Screen Shot 2019-03-10 at 10 51 14 PM](https://user-images.githubusercontent.com/5017156/54092109-7e97c680-4388-11e9-9867-f930e66d5958.png)

divyenduz:
![image](https://user-images.githubusercontent.com/746482/54747729-f4ddc980-4bcf-11e9-9a29-92832d56e8a2.png)

@garmjs : I am unable to reproduce, are you sure that the command that you are running doesn't yield an error status code or write to `stderr`? Can you please share your `prisma.yml` file or provide a minimal reproduction?

garmjs:
@divyenduz I'm not using prisma generate should i use instead of `graphql get-schema -p prisma`, this is prisma.yml file

```
endpoint: ${env:PRISMA_ENDPOINT}
datamodel: datamodel.prisma
hooks:
  post-deploy:
    - graphql get-schema -p prisma
    - graphql codegen
```

divyenduz:
@garmjs : Yes, that is recommended, most likely this is a bug in `graphql-cli` and it writes to `stderr` or ends with a non-zero status code. 

Please try a command like `echo "Running post-deploy hook"` using the following `prisma.yml` file to confirm this fix. 

```
endpoint: ${env:PRISMA_ENDPOINT}
datamodel: datamodel.prisma
hooks:
  post-deploy:
    - echo "Running post-deploy hook"
```

if that works without any red lines, please let me know and I will close the issue :) 

Please feel free to open a separate issue in `graphql-cli` package.

garmjs:
No red warning when I remove `graphql get-schema -p prisma`,  is there a way to get schema after deploy using Prisma CLI

divyenduz:
[...]

PR #3994, opened 2019-02-05
fix(deploy): do not write to stderr in non-error condition

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3977

### DIFF
--- a/cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
+++ b/cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
@@ -320,11 +320,11 @@ ${chalk.gray(
       const isPackaged = fs.existsSync('/snapshot')
       const spawnPath = isPackaged ? nativeSpawnSync : spawnSync
       const child = spawnPath(splittedHook[0], splittedHook.slice(1))
-      const stderr = child.stderr && child.stderr.toString()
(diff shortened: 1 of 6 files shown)


cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
```diff
@@ -320,11 +320,11 @@ ${chalk.gray(
       const isPackaged = fs.existsSync('/snapshot')
       const spawnPath = isPackaged ? nativeSpawnSync : spawnSync
       const child = spawnPath(splittedHook[0], splittedHook.slice(1))
-      const stderr = child.stderr && child.stderr.toString()
+      const stderr = child.stderr && child.stderr.toString().trim()
```
timsuchanek: What's the purpose of the `.trim()`?

divyenduz: There were two empty spaces for one `stderr`, making the condition below this pass. Trim removes trailing spaces making this more reliable. 

Is there any downside of using `trim` that you can think of?

timsuchanek: It could just have unwanted side effects as this is affecting every command.
Let's merge it for now and observe in beta if anything unwanted happens.

Issue #3987, opened 2019-02-04
Go client adds additional quotes to `DefaultEndpoint`


nikolasburk:
When generating the Prisma client in Go, it generates the `DefaultEndpoint` like so:

```go
var DefaultEndpoint = ""https://eu1.prisma.sh/nikolas/prisma-test/ytr""
```

When the client is used then, Go throws this error:

```
prisma-client/prisma.go:66:25: syntax error: unexpected https after top level declaration
```

This can be fixed by removing the extra pair of quotes:

```go
var DefaultEndpoint = "https://eu1.prisma.sh/nikolas/prisma-test/ytr"
```

You can reproduce this by following the instructions in any of the Go examples or by just using `prisma init` and generating the Go client.

bmschwartz:
It may be on older versions but I experienced this on 1.25.7

divyenduz:
This is available in the latest `stable` - `1.26.6`, please install it via `npm install -g prisma`.

PR #3996, opened 2019-02-05
fix(go-client): do not call replaceEnv twice

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3987

`replaceEnv` was being called twice due to legacy code.

Since, `replaceEnv` is called in `generate.ts` (generate command), the generated snapshot of `render` function of go client will have `DefaultEndpoint` without quotes which looks like incorrect Go code.

This actually unifies the behavior of all clients, if we test any other client with `endpoint` in options. It will yield code without quotes.


divyenduz:
Closing in favor of https://github.com/prisma/prisma/pull/4009

Issue #3359, opened 2018-10-24
`graphql-schema` generator should output a single file


nikolasburk:
The built-in generator for `graphql-schema` in the Prisma CLI should output a single file instead of a directory. 

Currently it's configured as follows:

```yml
generate:
    - generator: graphql-schema
      output: ./generated/prisma/
```

Instead, `output` should simply specify the name of a single file:

```yml
generate:
    - generator: graphql-schema
      output: ./generated/prisma.graphql
```

divyenduz:
This is available in the latest `alpha` - `1.28.0-alpha.13`, please install it via `npm install -g prisma@alpha`. Please be advised that this is a breaking change.

divyenduz:
This is available in the latest beta `1.29.0-beta`, please install it via `npm install -g prisma@beta`. Please be advised that this is a breaking change.

divyenduz:
This is available in the latest stable, `1.29.x`, please install it via `npm install -g prisma`. I am closing this issue, please reopen a new issue if required.

nunovieira:
I've come across this breaking change.
The documentation needs to be updated. Is that being done or should I try to submit a PR?
I also didn't notice any mention of this in the release notes.

pantharshit00:
@nunovieira I have updated the release notes.

PR #4080, opened 2019-02-21
[BREAKING] fix(generate): emit GraphQL schema to a single file

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3359


divyenduz:
This is available in the latest `alpha` - `1.28.0-alpha.13`, please install it via `npm install -g prisma@alpha`. Please be advised that this is a breaking change.

PR #4089, opened 2019-02-22
Update prisma-basics-datamodel-client-and-server-fgz4.mdx

manatran:
Fixed a simple typo.


pantharshit00:
Hi @manatran 

Thanks for catching this 🙌 

Can you please make this change in v1.27 docs as well? Then I can happily merge this.

PR #4113, opened 2019-02-28
models cleanup

mavilein:
This PR does the following things:
* it removes obsolete fields that we won't need anymore
* schema must not have project as a parent because we have cases where we have a schema but not a project (in the migration engine)
* come up with a suggestion to ease working with parent child relationships (see my comment in the changed files for an example)

### DIFF
--- a/server/prisma-rs/prisma-models/src/relation.rs
+++ b/server/prisma-rs/prisma-models/src/relation.rs
@@ -137,8 +137,8 @@ impl Relation {
     const MODEL_B_DEFAULT_COLUMN: &'static str = "B";
 
     fn schema(&self) -> SchemaRef {
-        self.schema.upgrade().unwrap_or(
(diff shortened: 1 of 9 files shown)


server/prisma-rs/prisma-models/src/relation.rs
```diff
@@ -104,7 +104,7 @@ impl RelationTemplate {
         };
 
         let model_a = {
-            let model = relation.with_schema(|schema| schema.find_model(&model_a_name).unwrap());
+            let model = relation.schema().find_model(&model_a_name).unwrap();
```
mavilein: I think this is easier to read than the version where you have to provide a function.

pimeys: It is, was planning to remove these tomorrow...

PR #4137, opened 2019-03-06
Index Comments/Mark as Error

ejoebstl:
Invalid indices can now be marked with a comment. This is used when one of the fields has an error.


ejoebstl:
@timsuchanek 
We should still merge this, if possible. Adding comments to indices might prove useful in the future.

PR #4162, opened 2019-03-11
Add prisma admin routes

dpetrick:
Adds path suffix `_admin` to access Prisma Admin on `/service/stage`. E.g.
- `/_admin` (implicit `/default/default/_admin`)
- `/service/_admin` (implicit `/service/default/_admin`)
- `/service/name/_admin`

are all valid paths to access Prisma Admin.

### DIFF
--- a/server/libs/sangria-server/src/main/resources/admin.html
+++ b/server/libs/sangria-server/src/main/resources/admin.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <meta name="theme-color" content="#000000" />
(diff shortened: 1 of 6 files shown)


server/libs/sangria-server/src/main/resources/admin.html
```diff
@@ -0,0 +1,40 @@
...
+    <script src="https://cdn.jsdelivr.net/npm/@prisma/prisma-admin/dist/app.js"></script>
+    <title>Prisma Admin</title>
+</head>
+<body>
+<noscript> You need to enable JavaScript to run this app. </noscript>
+<div id="root"></div>
+</body>
+<script>
+      window.addEventListener('load', function (event) {
+        PrismaAdmin.init(root, {})
```
Huvik: Could you please add `singleProject: true` so user could only use one project.
```suggestion
        PrismaAdmin.init(root, {singleProject: true})
```

PR #4213, opened 2019-03-19
Update authentication-and-security-kke4.mdx

lunchboxer:
made instructions for securing management api more explicit to include use of docker-compose up to effect changes and using enviroment variable to authenticate from command line.


nikolasburk:
Thanks a lot for the PR @lunchboxer, since we have added a new version of the docs in the meantime, I'll manually add your changes to that version as well 🙌

Issue #4226, opened 2019-03-20
adding a Node type breaks prisma cli generation and prisma server


pantharshit00:
**Describe the bug**
Adding a Node type to the datamodel breaks the prisma API as it already has a node type. It also breaks the cli generation.

**To Reproduce**
Try to generate a client or deploying and querying the data from the following datamodel:
```
enum Gender {
  MALE
  FEMALE
  DIVERSE
}

type User {
  id: ID! @unique
  username: String! @unique
  email: String! @unique
  password: String!
  verified: Boolean! @default(value: "false")
  verificationKey: String
  gender: Gender
  website: String
  bio: String
  registered: DateTime!
  sessions: [Session!]!
  gameServers: [GameServer!]!
}

type Session {
  id: ID! @unique
  user: User!
  cookie: String!
  remember: Boolean!
  browser: String
  os: String
  device: String
  lastActivity: DateTime
  expiresAt: DateTime!
  createdAt: DateTime!
}

type Node {
  id: ID! @unique
  location: NodeLocation!
  name: String
  description: String
  public: Boolean!
  fqdn: String!
  daemonSecret: String! @unique
  daemonHttps: Boolean!
  daemonPort: Int!
  daemonSftpPort: Int!
  cpuFrequency: Float! # in GHz
  cpuThreads: Float!
  ram: Int! # in mb
  diskspace: Int! # in gb
  ips: [NodeIP!]!
  games: [NodeGameAllocation!]!
  gameServers: [GameServer!]!
}

type NodeLocation {
  id: ID! @unique
  code: String!
  name: String!
  nodes: [Node!]!
}

type NodeIP {
  id: ID @unique
  node: Node!
  ip: String!
}

type Game {
  id: ID! @unique
  short: String! @unique
  name: String! @unique
  dockerImage: String!
  startCommand: String!
  stopCommand: String
  stopTimeout: Int
  minCpuThreads: Float!
  maxCpuThreads: Float!
  cpuPrice: Float! # in € per thread and GHz
  minRam: Int! # in mb
  maxRam: Int! # in mb
  ramPrice: Int! # in € per 1024mb
  minDiskspace: Int!
  maxDiskspace: Int!
  diskspacePrice: Int! # in € per gb
  gamedigType: String
  gamedigPortName: String
  description: String
  steamAppId: Int
  installLocalCopy: Boolean!
  latestVersion: String
  consoleCommandsEnabled: Boolean!
  crashProtectionEnabled: Boolean!
[...]

PR #4262, opened 2019-03-26
Validate use of reserved type names in datamodel

do4gr:
* add a datamodel validation for type names Prisma uses in the schema
* these are: `Mutation`, `Query`, `Subscription`, `Node`, `PageInfo`, `BatchPayload`
* Additionally users could also face errors if they define a typeName and a second typeName that equals a name Prisma would derive from the first i.e. `A` and `ACreateInput`
Fixes https://github.com/prisma/prisma/issues/4226

### DIFF
--- a/server/servers/deploy/src/main/scala/com/prisma/deploy/migration/validation/DataModelValidatorImpl.scala
+++ b/server/servers/deploy/src/main/scala/com/prisma/deploy/migration/validation/DataModelValidatorImpl.scala
@@ -322,7 +322,8 @@ case class ModelValidator(doc: Document, objectType: ObjectTypeDefinition, capab
       tryValidation(validateMissingTypes),
       tryValidation(requiredIdDirectiveValidation.toVector),
       tryValidation(validateRelationFields),
-      tryValidation(validateDuplicateFields)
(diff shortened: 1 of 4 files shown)


agonbina:
This should have been a major release since it is a breaking change isn't it?

do4gr:
I see this as a bug fix since using any of these names would have resulted in run-time errors upon first loading the schema from the server.

agonbina:
Yes and no ... Say I was using the previous version and (for some strange reason) I had `type Mutation { ... }` in my data model file. 

- I now upgraded the prisma server to this version (because it is just a patch)
- The server has an existing prisma service that was using one of the reserved types
- The server loads the schema for this existing service and will start throwing exceptions because the existing schema is using that type 
- All client relying on this prisma service are now rendered unusable (even though there was no changes to the prisma service itself)

do4gr:
Not sure I can follow your argumentation. 

My understanding is that if your server on the previous version had a reserved name as a type in your data model file it would have errorred immediately anyways and would have been unusable. So this is not breaking existing working servers. It changes the behaviour of (hypothetical) existing broken servers, which would now get a proper error instead of just a runtime error.

Issue #2306, opened 2018-04-29
Ability to disable hooks for an environment


FredyC:
# Feature Request

**What feature are you missing?**

The barely documented `hooks` in `prisma.yml` are always executed despite the environment. For example, When deploying to a local server, I get my graphql schema and bindings updated. However, I want to avoid running the same thing when deploying to other environments. It's just unnecessary holdup even though it correctly identifies there are no changes.

**How could this feature look like in detail? Tradeoffs?**

It would probably require an expanded structure than the current one so the flag to disable can be added, eg. based on environment variable can be specified.

However, I think that these hooks shouldn't be in `prisma.yml` at all. It feels like an ad-hoc solution as there are no other hooks at the moment. Instead of removing the automatic execution of this generation process, it would make much more sense to have a flat boolean variable for disabling it if anyone needs it. I am not sure why this was removed in the first place, perhaps to increase customization level, but I think both solutions can live peacefully together and for anyone who doesn't need customize it, there won't be anything extra to be done.

FredyC:
One more reason to actually handle this differently. If the service is secured, it's necessary to generate token and use it within `.graphqlconfig`. Attempting to generate schema from a different environment fails because token is wrong.

```yml
  database:
    schemaPath: src/generated/prisma.graphql
    extensions:
      prisma: src/database/prisma.yml
      prepare-binding:
        output: src/generated/prisma.ts
        generator: prisma-ts
      endpoints:
        default:
          url: "ENDPOINT"
          headers:
            Authorization: "TOKEN"
```

marktani:
> The barely documented hooks in prisma.yml

Thanks for your feedback. The [upgrade guide to 1.7](https://www.prisma.io/docs/1.7/reference/upgrade-guides/upgrading-prisma/upgrade-to-1.7-iquaecuj6b) includes explanations and examples for post deployment hooks. We'll soon add general documentation about post deployment hooks, check [this](https://github.com/graphcool/prisma/issues/2311).

> are always executed despite the environment

Yup, that's right! we should investigate how to improve this feature to allow for conditional or environment specific hooks.

> However, I think that these hooks shouldn't be in prisma.yml at all. It feels like an ad-hoc solution as there are no other hooks at the moment.

Where do you suggest these hooks to live instead?

> Instead of removing the automatic execution of this generation process [...] I am not sure why this was removed in the first place

[...]

PR #4317, opened 2019-04-05
feat(disable hooks): via a new flag

divyenduz:
Implements
https://github.com/divyenduz/rfcs/blob/disable-hooks/text/0000-disable-hooks.md

Closes https://github.com/prisma/prisma/issues/2306

Does not implement the `--hook` part of spec as that is redundant
(subjected to further discussion)

### DIFF
--- a/cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
+++ b/cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
@@ -65,6 +65,9 @@ ${chalk.gray(
     'no-generate': flags.boolean({
       description: 'Disable implicit client generation',
     }),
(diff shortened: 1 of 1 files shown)


cli/packages/prisma-cli-core/src/commands/deploy/deploy.ts
```diff
@@ -382,7 +392,7 @@ ${chalk.gray(
         !this.flags['no-seed'] &&
         projectNew
       ) {
-        this.printHooks()
+        this.printHooks(noHook)
```
timsuchanek: It's a taste question, but speaks something against just having an `if(nooHook)` here instead of introducing an arg to `printHooks`?

Issue #3547, opened 2018-11-19
CLI performance


mxstbr:
**Describe the bug**
The CLI is notably slow to do anything, taking > 1s to start logging the first output.

**To Reproduce**

Run a simple `--version` call and note how long it takes:

```
$ time prisma --version
...
0.85s user 0.13s system 2.064 total
```

In comparison:

```sh
$ time grep --version
...
0.00s user 0.00s system 0.008 total

# and a little bit more fairly
$ time npm --version
...
0.18s user 0.05s system 0.272 total
```

Grep is logs almost instantly, but even npm (which is also written in JS) is 10x faster. 

**Expected behavior**
The CLI logs the first output quickly

schickling:
Thanks a lot for reporting this. We expect this to massively improve via #1480. 🙏

mxstbr:
Are you sure that's the bottleneck when running commands? I can't imagine Node taking > 1s just to resolve the CLIs dependencies, there must be something else happening that's taking so long.

timsuchanek:
@mxstbr we have the architecture (don't use it yet) for an fs-based plugin system. These plugins are scanned in the cli on bootup time. Right now I assume this is the biggest bottleneck.

PR #4319, opened 2019-04-05
feat(warning): when server and CLI are not in sync

divyenduz:
Implements
https://github.com/divyenduz/rfcs/blob/cli-server-sync/text/0000-cli-server-sync.md

Two concerns that need to be reviewed: 

1. This pollutes the `Command` class to be aware of cluster (by initiating one), which might be ok since some commands now rely on getting actual server version. 

1. `await cmd.definition.load(cmd.flags, envFile)` in `areServerAndCLIInSync ` would load the environment twice for some command, making them further slow https://github.com/prisma/prisma/issues/3547

### DIFF
--- a/cli/packages/prisma-cli-engine/src/Command.ts
+++ b/cli/packages/prisma-cli-engine/src/Command.ts
@@ -12,6 +12,10 @@ import * as mock from './mock'
 import { RC } from './types/rc'
 import { initStatusChecker } from './StatusChecker'
 import { filterObject } from './util'
+import * as os from 'os'
+import * as fs from 'fs-extra'
+import * as path from 'path'
+import chalk from 'chalk';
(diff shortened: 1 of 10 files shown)


cli/packages/prisma-cli-engine/src/Command.ts
```diff
@@ -56,11 +61,93 @@ export class Command {
...
+    }
+  }
+
   static async run(config?: RunOptions): Promise<Command> {
     const cmd = new this({ config })
 
     try {
       await cmd.init(config)
+      const { inSync, serverVersion } = await cmd.areServerAndCLIInSync(cmd)
+      if (this.printVersionSyncWarning && !inSync) {
```
timsuchanek: Speaks anything against switching these 2 lines out?

Issue #4020, opened 2019-02-07
CLI output for post-deploy hooks is a bit confusing


nikolasburk:
Asume I have these post-deploy hooks in my prisma.yml:

```yml
hooks:
  post-deploy:
    - prisma generate
    - npx nexus-prisma-generate ./src/generated/prisma-client ./src/generated/nexus-prisma 
```

When invoked, the CLI output will be similar to this:

```
Running prisma generate...
Saving Prisma Client (TypeScript) at /Users/nikolasburk/prisma/github/prisma-examples/typescript/graphql/src/generated/prisma-client/
Running prisma generate ✔

Running npx nexus-prisma-generate ./src/generated/prisma-client ./src/generated/nexus-prisma...
ERROR: Missing or invalid argument --client
Running npx nexus-prisma-generate ./src/generated/prisma-client ./src/generated/nexus-prisma ✖
```

In general I think the idea of having start- and end-"checkpoints" make a lot of sense. But the way it's laid out at the moment is a bit confusing because the important information (i.e. the `...` and the ` ✔`) are at the _end_ of the lines, the beginning and the rest of the line look exactly the same. For the longest time I've thought this was a bug that would print the command twice until I realized that the end was different.

So my suggestion is to adjust the wording of the checkpoints to make it clear from the beginning what these messages are about.

nikolasburk:
I'd suggest the following wording and characters:

```
▶︎ Running post-deployment hook: prisma generate
Saving Prisma Client (TypeScript) at /Users/nikolasburk/prisma/github/prisma-examples/typescript/graphql/src/generated/prisma-client/
✔ Post-deployment hook succeeded: prisma generate

▶︎ Runing post-deployment hook: npx nexus-prisma-generate ./src/generated/prisma-client ./src/generated/nexus-prisma
ERROR: Missing or invalid argument --client
✖ Post-deployment hook failed: npx nexus-prisma-generate ./src/generated/prisma-client ./src/generated/nexus-prisma
```

PR #4346, opened 2019-04-10
feat(cli-wording): adjust some wording

divyenduz:
Partial https://github.com/prisma/prisma/issues/4020


cli/packages/prisma-cli-core/src/commands/import/index.ts
```diff
@@ -33,8 +34,17 @@ export default class Import extends Command {
     if (
       this.definition.definition!.databaseType &&
       this.definition.definition!.databaseType === 'document'
-    ) {
-      throw new Error('Import is not yet supported for document stores.')
+      ) {
+        throw new Error(`Import is not yet supported for document stores. Please use the native export features of your database.
```
nikolasburk: Looks good, but it says: "Please use the native **export** features" when it's for the `import` command.

Issue #3723, opened 2018-12-18
sync-exec; 1 sync-exec vulnerability found in package-lock.json


steida:
```
npm list sync-exec
este@ /Users/steida/dev/este-typescript
└─┬ prisma@1.22.2
  └─┬ prisma-cli-core@1.22.2
    └─┬ npm-run@4.1.2
      └── sync-exec@0.6.2
```

Should not be used imho.

divyenduz:
There is a regression in windows for `npm-run` (https://github.com/prisma/prisma/pull/3517). We can upgrade this after this one is merged https://github.com/timoxley/npm-run/pull/21

Or break the build for windows temporarily. Can you please confirm what the vulnerability is exactly and does it also affect CLI tools?

steida:
I think it's not severe, because it's for local web development, but I am not an expert. I would wait, this issue can help the others meanwhile.

tianhuil:
The regression is still around.  Any ETA on fixing this?
```
$ npm list sync-exec
project@0.0.3
└─┬ prisma@1.34.1
  └─┬ prisma-cli-core@1.34.1
    └─┬ npm-run@4.1.2
      └── sync-exec@0.6.2
```

PR #4373, opened 2019-04-13
Upgrade `npm-run`

tianhuil:
Avoid a vulnerability in older version of `npm-run`'s `sync-exec@0.6.2`: https://nvd.nist.gov/vuln/detail/CVE-2017-16024

### DIFF
--- a/cli/packages/prisma-cli-core/package.json
+++ b/cli/packages/prisma-cli-core/package.json
@@ -101,7 +101,7 @@
     "multimatch": "^2.1.0",
     "mysql": "^2.16.0",
     "node-forge": "^0.7.1",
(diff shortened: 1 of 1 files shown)


pantharshit00:
Hi @tianhuil 

As discussed in https://github.com/prisma/prisma/issues/3723#issuecomment-448143954 upgrading it to v5 will break windows support. RIght now we are waiting for forhttps://github.com/timoxley/npm-run/pull/21. But this vulnerability only occurs in development so I think we are safe waiting rather than breaking windows build.

We are already planning for the next iteration of our CLI for prisma 2 and in that, we will avoid using this altogether.

PR #4387, opened 2019-04-16
Split DatabaseWrite into separate traits

pimeys:


pimeys:
Refactoring round. Splitting stuff into traits again and doing some tests and commentary.

Issue #3774, opened 2018-12-27
Prisma client: incorrect TS types for nullable field


dodas:
Hey guys.
I just noticed that generated TS types for `SomeTypeUpdateInput` doesn't allow `null` for fields that are nullable.
E.g.:
datamodel.graphql:
```graphql
type User {
  id: ID! @unique
  nickname: String
}
```

Trying to update `nickname` to null:
```js
prisma.updateUser({
  where: { id: 'abcdef' },
  data: { nickname: null },
});
```
Executing this code works and have desired outcome, however TypeScript complains:
```
Type 'null' is not assignable to type 'string | undefined'
```

I guess this is bug in generated types?
Thanks.

dodas:
I updated both `prisma-client-lib` and global `prisma` cli to latest versions, regenerated client, restarted TS language server in VSCode and still having same issue.
I am not an TS expert, but I guess that when interface says `String` (which is just alias of `string`), it cannot accept `null`.

pantharshit00:
No that is not the case. The `?` operator makes it nullable. 

If your project is open source or you can put together a minimal reproduction, please share it.

dodas:
Okay, I think I found the reason why I am getting error and you not.
In your `tsconfig.json`, try setting `"strict": true` under `"compilerOptions"` and we should probably have same result.

Note that `?` operator makes an property _optional_, not _nullable_. That is, you can omit it altogether, but not pass `null` or `undefined`.

So I think prisma should generate nullable types as `String | null` to allow strict mode.

pantharshit00:
Thanks for the reply!


Main reason here is strictNullCheck and how TS handles null differently when this option is enabled.

You can omit the nickname property for now to achieve the null behavior(ie it will be undefined but we will write it as null in db). We need a bit more discussion internally for inclusion of this change in code generation.

dodas:
I just tried ommiting an nullable property and it doesn't update it to null - and honestly, that would be very surprising and dangerous behaviour.
Until this is resolved, I am just using `// @ts-ignore` before line that is passing `null`.

pantharshit00:
I though you were using it in `create`.

Indeed this needs a proper resolution now.

MathiasKandelborg:
Hi! I've been trying to solve this via graphqlgen. [Here's the issue on that repo](https://github.com/prisma/graphqlgen/issues/417)

[...]

PR #4449, opened 2019-04-25
fix(ts-client): lenient input optionals

divyenduz:
Fixes https://github.com/prisma/prisma/issues/3774

Makes the optional inputs more lenient and allows null to be inputted
for an input type. Making it play better with typescript's strict mode.

Open question:

Does sending null for an optional input field work well with the
runtime?
- Yes it does, tested

This is not a classic type system `Maybe` type and we might want to change the name to something else.


P4sca1:
Could you explain the difference in assigning nothing (`undefined`) vs assigning `null` to an optional prisma argument?

divyenduz:
In Javascript, undefined means an `uninitialized` variable while `null` is an initialized variable but there is no value to fulfill it. Prisma input API would treat these as the same i.e. `null`. It was more strict before but now we are more lenient.

PR #4470, opened 2019-04-29
Docs: fix docker-compose yml volume declarations

leonardodino:
`docker-compose.yml` uses implicit block values for volumes, prettier seems to have some trouble with that.

This **`PR`** adds a yml null character (`~`) for the value of the volume.
This is the [recommended solution](https://github.com/prettier/prettier/issues/4971#issuecomment-412257910) in the upstream [prettier issue](https://github.com/prettier/prettier/issues/4971).

---

Another approach was [adding a blank line](https://github.com/leonardodino/prisma/commit/c7f35d2ac01b182d4651c3079c2014d7d2c8ed9e) to the mdx yaml block.

**upside**: the `docs` would reflect how "null" value volumes are described at `docker-compose` documentation. (by leaving the line empty)

**downside**: running prettier wouldn't be idempotent:
- the first time it runs, it removes the trailing newline but keeps the format as desired;
- the next runs will break the formatting, as the newline won't be there anymore.

this wouldn't be a problem in the CI servers, where it's always a clean slate,
but it could be a problem when running on a dev machine.

---

fixes: https://github.com/prisma/prisma-content-feedback/issues/89
related: https://github.com/prettier/prettier/issues/4971

### DIFF
--- a/docs/1.0/03-Tutorials/04-Cluster-Deployment/02-Digital-Ocean-(Docker-Machine).md
+++ b/docs/1.0/03-Tutorials/04-Cluster-Deployment/02-Digital-Ocean-(Docker-Machine).md
@@ -145,7 +145,7 @@ networks:
     driver: bridge
 
 volumes:
-  db-persistence:
+  db-persistence: ~
 ```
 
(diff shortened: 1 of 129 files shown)


nikolasburk:
Thanks a lot for the fix Leo 🙌

PR #4511, opened 2019-05-08
Add repository to package.json

guerrerocarlos:


cli/packages/prisma-client-lib/package.json
```diff
@@ -6,8 +6,10 @@
   "files": [
     "dist"
   ],
+  "repository": "prisma/prisma",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
+  https://github.com/prisma/prisma/tree/master/cli/packages/prisma-client-lib
```
leonardodino: this line shouldn't be here
```suggestion
```

cli/packages/prisma-client-lib/package.json
```diff
@@ -6,8 +6,10 @@
   "files": [
     "dist"
   ],
+  "repository": "prisma/prisma",
```
leonardodino: consider using the recently released [monorepo declaration](https://github.com/npm/rfcs/blob/latest/implemented/0010-monorepo-subdirectory-declaration.md), so it can point to each package inside this repo (:

PR #4529, opened 2019-05-13
If field behaviour is `id`, the identifier is GraphQLID

pimeys:


server/prisma-rs/prisma-models/src/node.rs
```diff
@@ -101,7 +101,10 @@ impl Node {
             })?;
 
         Ok(match &self.values[index] {
-            PrismaValue::GraphqlId(ref id) => id,
+            PrismaValue::GraphqlId(ref id) => id.clone(),
```
dpetrick: Maybe TryFrom?

PR #4563, opened 2019-05-17
Start optimizing filters right after parsing them

do4gr:


server/connectors/api-connector-mongo/src/main/scala/com/prisma/api/connector/mongo/database/AggregationQueryBuilder.scala
```diff
@@ -79,9 +79,10 @@ trait AggregationQueryBuilder extends FilterConditionBuilder with ProjectionBuil
       //-------------------------------RECURSION------------------------------------
       case NodeSubscriptionFilter => Seq(`match`(hackForTrue))
       case AndFilter(filters)     => sortFilters(filters).flatMap(f => buildJoinStagesForFilter(path, f))
-      case OrFilter(_)            => sys.error("This is not implemented for the Mongo connector")
-      case NotFilter(_)           => sys.error("This is not implemented for the Mongo connector")
-      case x: RelationFilter      => relationFilterJoinStage(path, x)
+      case OrFilter(filters) if filters.forall(f => f.isInstanceOf[ScalarFilter]) =>
+        Seq(`match`(or(filters.map(f => buildConditionForScalarFilter(path.combinedNames, Some(f))): _*)))
```
mavilein: what's that? :D

server/servers/api/src/main/scala/com/prisma/api/schema/Optimizations.scala
```diff
@@ -0,0 +1,126 @@
...
+    def optimize(filter: Filter): Filter
+
+  }
+
+  object FilterOptimizer extends Optimizer {
+
+    override def optimize(filter: Filter): Filter = {
+      val one = LogicalOpt.transform(filter)
+//      val two   = InlineOpt.transform(one)
+      val three = SameRelationFilterOpt.transform(one)
```
mavilein: To avoid naming all those intermediate variables you could put all the `.transform` functions into an an array and use a `foldLeft` to apply all of them.

server/servers/api/src/main/scala/com/prisma/api/schema/Optimizations.scala
```diff
@@ -0,0 +1,126 @@
...
+      three
+    }
+
+    trait Optimization {
+      def transform(filter: Filter): Filter
+    }
+
+    object LogicalOpt extends Optimization {
```
mavilein: I think it would be beneficial to add some explanations of what this filter does. Right now i don't understand the name. From reading it looks to me as if this removes superfluous wrapping of filters. But still i am not sure. This topic seems so involved an explanation is warranted :D

server/servers/api/src/main/scala/com/prisma/api/schema/Optimizations.scala
do4gr: done

server/servers/api/src/main/scala/com/prisma/api/schema/Optimizations.scala
do4gr: That's what it does. Added a comment.

server/connectors/api-connector-mongo/src/main/scala/com/prisma/api/connector/mongo/database/AggregationQueryBuilder.scala
do4gr: I'll add a comment ;-)

PR #4587, opened 2019-05-27
Query Schema builder

dpetrick:


dpetrick:
Also merges https://github.com/prisma/prisma/pull/4607

PR #4598, opened 2019-05-29
Datamodel relations

ejoebstl:
RS/datamodel:
    * Basic relation validation
    * Auto-Add backwards relation field


server/prisma-rs/libs/datamodel/src/dml/field.rs
```diff
@@ -62,6 +62,9 @@ pub struct Field {
     pub scalar_list_strategy: Option<ScalarListStrategy>,
     /// Comments associated with this field.
     pub comments: Vec<Comment>,
+    /// If set, signals that this field was internally generated
+    /// and should never be displayed to the user.
+    pub generated: bool,
```
ejoebstl: I've added a `generated` flag to identify generated fields.

server/prisma-rs/libs/datamodel/src/dml/validator/mod.rs
```diff
@@ -98,13 +127,213 @@ impl Validator {
...
+            }
+        }
+
+        Ok(())
+    }
+
+    /// For any relations which are missing to_fields, sets them to the @id fields
+    /// of the foreign model.
+    #[allow(unused)] // No normalization of to_fields for now.
+    fn set_relation_to_field_to_id_if_missing(&self, schema: &mut dml::Schema) {
```
ejoebstl: I've left this in, in case we need it later. It's not called anywhere.

server/prisma-rs/libs/datamodel/src/dml/validator/mod.rs
```diff
@@ -98,13 +127,213 @@ impl Validator {
...
+        for model in schema.models() {
+            missing_back_relations.append(&mut self.find_fields_with_missing_back_relation(model, schema));
+        }
+
+        for (forward, backward) in missing_back_relations {
+            let model = schema.find_model_mut(&forward.to).expect(STATE_ERROR);
+
+            // TODO: Ugly generated name for now.
+            model.add_field(dml::Field::new_generated(
+                &format!("generated_{}", backward.to),
```
ejoebstl: I would expect the name of the generated back relation field to not be important. Anyway, I'm very open to changing that.

mavilein: It is actually important. That field will also be visible in the client. We want to user the camelCased variant of the name of the related type in this case.

PR #4604, opened 2019-05-30
Datamodel Rendering and Spec Consistency

ejoebstl:
* Added basic rendering of Datamodel ASTs
* Added conversion from Datamodel to AST (with extensiblity!)
* Back relation fields are now optional
* Removes support for `Post(id)` related fields notation
* Changed block directive notation to `@@` inside block
* Top level rendering/parsing helpers
* DMMF update to newest spec


server/prisma-rs/libs/datamodel/src/lib.rs
```diff
@@ -64,10 +66,42 @@ pub fn parse_with_plugins(
     validator.validate(&ast)
 }
 
+/// Parses and validates a datamodel string, using core attributes only.
 pub fn parse(datamodel_string: &str) -> Result<Datamodel, errors::ErrorCollection> {
     return parse_with_plugins(datamodel_string, vec![]);
 }
 
+/// Parses a datamodel string to an AST. For internal use only.
```
ejoebstl: Here are the top-level helpers.

server/prisma-rs/libs/datamodel/tests/parser_renderer.rs
```diff
@@ -0,0 +1,71 @@
...
+#[test]
+fn test_parser_renderer_via_ast() {
+    let ast = datamodel::parse_to_ast(DATAMODEL_STRING).unwrap();
+    let rendered = datamodel::render_ast(&ast);
+
+    assert_eq!(DATAMODEL_STRING, rendered);
+}
+
+#[test]
+fn test_parser_renderer_via_dml() {
```
ejoebstl: This test is probably a good high-level usage example.

PR #4610, opened 2019-06-01
Add prisma introspect command

Weakky:
One concern:

That PR adds a *lot* of dependencies because of the CLI. It might be wise to keep the cli separated from the actual runtime introspection code


Weakky:
Not sure yet why the tests are failing. It seems like `ink` have conflicting types with react, but `skipLibCheck` is enabled on the tsconfig.json and should therefore skip these errors

PR #4618, opened 2019-06-03
Builtin-Source Config Blocks

ejoebstl:
For mysql, postgres, sqlite.

There are no special config keys defined, but the base structure is there.


server/prisma-rs/libs/datamodel/tests/config/builtin.rs
```diff
@@ -0,0 +1,80 @@
+use crate::common::ErrorAsserts;
+use datamodel::errors::ValidationError;
+
+const DATAMODEL: &str = r#"
+source pg1 {
```
ejoebstl: Have a look at this test for an example.

PR #4688, opened 2019-06-21
Fix incorrect env var reference to PRISMA_MANAGEMENT_API_SECRET

realAlexBarge:
This file at one point incorrectly refers to the needed PRISMA_MANAGEMENT_API_SECRET env var as PRISMA_MANAGEMENT_SECRET


berniedurfee:
Thank you @realAlexBarge!! This cost me some time. :)

PR #4709, opened 2019-06-30
Go client: implement relay connections

### LINKED ISSUE #4027: Go Client: Implement Relay Connections
**Describe the bug**
Connections are currently not supported in the Go client. Some types already exist (Edge, PageInfo...), but the functions are not implemented yet.

**To Reproduce**
Create a single-to-many relation between User and Post:
```graphql
type User {
	id: ID! @unique
	email: String! @unique
	name: String!
	posts: [Post!]!
}

type Post {
	id: ID! @unique
	user: User!
	title: String!
	content: String!
}
```

Generate the Prisma client and access it:
```go
func (r *queryResolver) Feed(ctx context.Context) ([]prisma.Post, error) {
	conn, err := r.Prisma.PostsConnection(&prisma.PostsConnectionParams{}).Exec(ctx)
	// conn is a struct with no further methods and PostsConnections panics
}
```

**Next steps**

An actual implementation is missing for the connection methods. Also, some structs are empty.

```
func (client *Client) PostsConnection(params *PostsConnectionParams) PostConnectionExec {
	panic("not implemented")
}

...

type PostConnection struct {
}
```

related: https://github.com/prisma/prisma/blob/214389359b94de9abec44e43110b70df46d2046a/cli/packages/prisma-client-lib/src/codegen/generators/go-client.ts#L588

steebchen:
Implement basic connections which return edges & pageInfo and fix
existing bugs/incorrect code.
The user can also query for edges or pageInfo separately.

This is a simple solution, taking into account that this codebase for prisma v1 is not maintained anymore and soon deprecated by prisma v2.

Fix #4027

Notes:
The connection return struct (e.g. `UsersConnection`) contains `edges` and `pageInfo`, which is in my opinion the most useful implementation. Always fetching `aggregate` (which currently contains `count` only) seems unnecessary, and you can still aggregate by doing `agg, err := prisma.Connection(...).Aggregate(ctx)`.


matthewmueller:
Looks good, thanks Luca!

steebchen:
It seems like the build failed (well, since you don't have CI running in PRs...):
`codegen › generators › go-client › go generator Did not match snapshot`
It seems like the snapshot update added some other fields (for whatever reason):

```
+ UserOrderByInputCreatedAtAsc UserOrderByInput = "createdAt_ASC"␊
+ UserOrderByInputCreatedAtDesc UserOrderByInput = "createdAt_DESC"␊
+ UserOrderByInputUpdatedAtAsc UserOrderByInput = "updatedAt_ASC"␊
+ UserOrderByInputUpdatedAtDesc UserOrderByInput = "updatedAt_DESC"␊
```

I can send a PR to fix this, but I don't understand why the diff was added in the first place.

cli/packages/prisma-client-lib/src/codegen/generators/go-client.ts
```diff
@@ -267,7 +267,37 @@ export class GoGenerator extends Generator {
...
+                      [3]string{"${objectName}WhereInput", "${objectName}OrderByInput", "${objectName}Edge"},
+                      "edges",
+                      []string{"cursor"})
+
+                    nodes := edges.Client.GetMany(
+                      edges,
+                      nil,
+                      [3]string{"", "", "${objectName}"},
+                      "node",
+                      []string{"id", "createdAt", "updatedAt", "name", "desc"})
```
chris-rock: I am not sure why name and desc are required table columns?

steebchen: This was by accident, it's fixed in #4730

chris-rock: Ah, thank you @steebchen for the explanation.

PR #4730, opened 2019-07-10
go client: fix relay connections

steebchen:
Fix some things introduced in #4709.

This looks like a big change but it's actually just a few changes solely related to go connections:

- use correct fields for edge->node selection
- use different Connection.Exec() method to return edges & pageInfo
- fix singular/plural naming with objectName

**Note**
The edges now successfully return an array of nodes, but I can't get it to work so that the cursor is returned as well, it's either the cursor or the node (and its fields), but it seems impossible to get both. I know you don't have much resources for the v1 Go client, but let me propose a few options and I'll take care of it:

1. Revert all go connections
2. Leave as is and accept that cursors are empty.
3. Export an .Edges() method where the cursor values are manually added (the cursor value equals to the node id anyway, so this can be easily constructed)
4. Introduce a .Nodes() method to just export nodes (will be useful for a lot of clients, the Github API does this as well), can be used in combiniation of 3 optionally

I say we go for 3 and 4 so that the Go client end-user won't notice anything but working stuff, namely they have access to .Nodes() and .Edges(). This is all non-breaking.

code line where edges/nodes are fetched:
https://github.com/prisma/prisma/pull/4730/commits/4eb62d64e6dcf28be00d98a31e99bfc76be35f84#diff-04c0cc2080a3ea69465276562ce0521bR295

cc @matthewmueller, can you please take a look at it?


steebchen:
After a bit more thinking, regardless of the implementation, I would just add a single public method `Connection(...)` which returns a result struct, partly because of the current limitations but mainly because Connections provide the `pageInfo` object.

```go
conn, err := r.Prisma.UserConnection(...)
handle(err)
log.Printf("conn: %+v", conn) // { PageInfo: { hasNext: true, ... }, Nodes: []User{{ ID: "123", Name: "John" }} }
```

That's it.

---

**Who says `nodes` is a thing?**
The relay spec only describes edges, but as I said previously edges is kind of useless for prisma v1. Nodes simplify things and it's a common practice to just return the nodes (for example, the Github GraphQL API returns nodes & edges).

**Why no `edges`?**
Due to the current Go client limitations, we can't properly fetch edges with the cursor and the node, which makes this cumbersome. Additionally, the cursor is quite useless because the cursor is just the node ID anyway, so we can just use `nodes` and we already have our cursor with `node.ID`.

PR #4742, opened 2019-07-15
Introspection now ignores _Migration table at database level.

ejoebstl:
Takes care of https://github.com/prisma/prisma2/issues/78.


ejoebstl:
(Force push to take care of `build-cli` module version which was diverged from alpha)

PR #4792, opened 2019-08-08
Database-introspection: Handle ON DELETE actions for foreign keys

aknuds1:
Handle `ON DELETE` actions for foreign keys in Postgres.


aknuds1:
I don't know why the tests break, locally they're all fine.

PR #4809, opened 2019-08-15
Make introspection schema serializable

aknuds1:
Make the schema returned by Rust-based introspection serializable.

### DIFF
--- a/server/prisma-rs/libs/database-introspection/tests/serialization_tests.rs
+++ b/server/prisma-rs/libs/database-introspection/tests/serialization_tests.rs
@@ -238,3 +238,47 @@ fn database_schema_is_serializable_for_every_column_type_family() {
     // Verify that schema deserialized from reference JSON is equivalent
     assert_eq!(ref_schema, schema);
 }
+
+#[test]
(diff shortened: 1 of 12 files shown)


aknuds1:
@mavilein Please let me know if the test is how you would want it. :)

server/prisma-rs/libs/database-introspection/tests/serialization_tests.rs
```diff
@@ -0,0 +1,147 @@
...
+        .level(log_level)
+        .chain(std::io::stdout())
+        .apply()
+        .expect("fern configuration");
+
+    IS_SETUP.store(true, Ordering::Relaxed);
+}
+
+#[test]
+fn database_schema_is_serializable() {
```
mavilein: This test is the direction i imagined. It would be great if we could add more cases though. The following things caught my eye:

- The test doesn't check whether null for a primary key works.
- The tests should catch when an enum variant is renamed. So we should exercise each enum variant we have.

aknuds1:
@mavilein Please have another look. I've tried to exhaust all enums with new test cases.

PR #4817, opened 2019-08-19
Add support for tabs in the datamodel

matthewmueller:
Currently the datamodel parser panics if you add tabs, this PR fixes it.

Closes: https://github.com/prisma/prisma2/issues/327


ejoebstl:
Yep, that fix looks good from my end.

Have you tried prisma-fmt with this fix & tabs?

matthewmueller:
Good call, just added tests for `reformat`. Seems to be working :-D

PR #4831, opened 2019-08-21
Fix rust warnings

aknuds1:
I got a lot of warnings after upgrading the Rust toolchain to 1.37, this PR fixes them via `cargo fix`. Also removed a couple of unused functions by hand, since `cargo fix` didn't take care of them.

### DIFF
--- a/server/prisma-rs/migration-engine/connectors/sql-migration-connector/src/database_inspector/mod.rs
+++ b/server/prisma-rs/migration-engine/connectors/sql-migration-connector/src/database_inspector/mod.rs
@@ -26,7 +26,7 @@ pub trait DatabaseInspector: Send + Sync + 'static {
     fn introspect(&self, schema: &String) -> DatabaseSchema;
 }
 
-impl DatabaseInspector {
+impl dyn DatabaseInspector {
     const PARSE_ERROR: &'static str = "Parsing of the provided connector url failed.";
 
     pub fn empty() -> EmptyDatabaseInspectorImpl {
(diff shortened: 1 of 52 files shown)


server/prisma-rs/migration-engine/connectors/sql-migration-connector/src/database_inspector/mod.rs
```diff
@@ -26,7 +26,7 @@ pub trait DatabaseInspector: Send + Sync + 'static {
     fn introspect(&self, schema: &String) -> DatabaseSchema;
 }
 
-impl DatabaseInspector {
```
pimeys: You could maybe move all of these methods directly to the trait.

aknuds1: @pimeys Could we do that in another PR? I tried, but it leads to non-trivial build failures.

pimeys: Can you show me what kind of failures? You're doing God's work here btw :)

aknuds1: @pimeys OK, I'll try to reproduce

aknuds1: @pimeys After discussing with Marcus, we'll merge this now and handle the trait in a follow-up PR. I'll get back to you later on this, OK? Got to handle another issue now.

pimeys:
Thank you :)

server/prisma-rs/migration-engine/connectors/sql-migration-connector/src/database_inspector/mod.rs
aknuds1: See https://github.com/prisma/prisma/pull/4836.

PR #4840, opened 2019-08-22
database-introspection: support hyphenated names

aknuds1:
Add test to ensure that hyphenated table/column names are supported. I also had to patch Barrel to make this work, now using the master branch of my fork of the latter.


aknuds1:
I'm just going to fix SQLite support for hyphenated schema name, since that breaks apparently.

Issue #4829, opened 2019-08-21
[Discussion] Streamlining Prisma's support channels


nikolasburk:
Hey everyone 👋 we want to use this GitHub issue to spark a discussion about the variety of Prisma's community and support channels and how to organize them better. Please leave your feedback and ideas!

We currently have the following official and unofficial community and support channels:

- Slack
- GitHub
- Prisma forum
- Spectrum
- Intercom
- Email
- Stackoverflow

Our support team is working very hard but taking care of all the channels at once is an almost impossible task. We're currently evaluating how to streamline the channels, here are a few of our thoughts.

While we currently use **Slack** as our main communication channel to the community, there are some problems Slack brings when used with a community as huge as ours (we currently have more than 18k members in our Slack). Mostly, there's no way to make Slack messages searchable on the web (e.g. via Google). This means we can't point to older replies when a question is being repeatedly asked on Slack and users don't have a good way of finding previous answers in the first place. The realtime aspect of Slack is great, but to provide sustainable support it should be possible to persist and reuse previous answers.

We use the **Prisma forum** to enable broader discussions among community members and respond to questions with our support team. These questions and answers can be persisted, but many of our users aren't aware of the forum and rather default to asking questions on Slack. Another issue is that our forum is running on discourse which requires our users to create an extra account.

**Spectrum** seemed to be a great middleground between the ability to provide realtime support and persisting conversations so that they could be reused later. Many companies have already moved their communities to Spectrum (e.g. [ZEIT](https://zeit.co/blog/spectrum) and [Apollo](https://blog.apollographql.com/goodbye-slack-hello-spectrum-8fa6b979645b)), however, we've been hesitant for a few reasons:

- We are worried about Spectrum's reliability as it has been facing regular downtimes
- Spectrum doesn't yet provide a mobile app which we consider an important factor
- The chat experience often is fairly slow

While we believe that the Spectrum team would be able to fix the above issues, we're worried that Spectrum isn't actively developed any more ([the last commits are from almost two months ago](https://github.com/withspectrum/spectrum/commits/alpha)). 

So, instead of moving our community to Spectrum, we'd rather hear your opinions and thoughts. Are you aware of a community management tool that might help or have other feedback to share? Please let us know your ideas!

gustawdaniel:
Is there any place when voting on features priorities is possible?

[...]

PR #4876, opened 2019-09-09
Change CLI output from spectrum to slack

pantharshit00:
See #4829


janpio:
Is the CI failure on the merge commit of this expected?

pantharshit00:
Yes, there is no tag that CI can use to publish the change

PR #4915, opened 2019-10-16
docs(readme): fix image height

steebchen:
This fixes the image being skewed on small screens

Before & after:

image

The only caveat is that it renders huge when comparing in split diff and rendered mode, which is probably okay:

image


janpio:
Using `max-height` _might_ fix the rendering problem thing that is not even relevant.

steebchen:
@janpio tried that already, didn't work :/

Issue #4417, opened 2019-04-19
Provide additional options for SSL MongoDB connections in configuration


kellerj:
**Describe the bug**
There are some options which can only be passed to the MongoDB driver via an options object.  (Not supported on the URI query string.) . Specifically, the ability to specify CA or key/cert files needed for connections to some MongoDB instances.  (And important for testing, the ability to turn off SSL certificate checking.) . They are the options that correspond to the mongo CLI options below:

--sslCAFile arg              
--sslPEMKeyFile arg          
--sslPEMKeyPassword arg      
--sslCRLFile arg             
--sslAllowInvalidCertificates

**To Reproduce**
1. Generate a self-signed cert/key and start up a mongo instance (`docker run -p 27018:27017 --name mongo --rm -v $PWD:/etc/ssl mongo:4 --sslMode requireSSL --sslPEMKeyFile /etc/ssl/mongodb.pem -v`)
2. Verify connections work to the database with the needed flags set (simplest case: `mongo localhost:27018 --ssl --sslAllowInvalidCertificates`)
3. Attempt to connect to this database using `mongodb://localhost:27018/?ssl=true` as the connect string with prisma

**Expected behavior**
An additional yml dictionary of connection options which could be passed along to the connection objects.

**Versions (please complete the following information):**
 - Connector: `MongoDB`
 - Prisma Server: `1.31.0`

kellerj:
Further information:

I continued hacking to see what I could come up with locally.  It looks like the forcing of netty by system property as in the files below may be triggering the error described here: https://jira.mongodb.org/browse/SCALA-501

Sounds like netty may no longer be needed for the MongoDB connections?

MongoDatabasesFactory.scala
MongoInternalDatabaseDefs.scala

EarthBug:
Is there any way to connect to a mongo instance which requires certificates currently?

cat-turner:
any update on this issue

larsivi:
Using scalegrid.io as the database provider, SSL is required for MongoDB. This functionality must be supported in prisma for the mongodb connector to make sense.

larsivi:
I have so far been using Mongo DB 3.4 (and upgrading isn't necessarily trivial), but it seems like 4.2 has added the possibility to specify the certificates, etc in the connection string, e.g. tlsCertificateKeyFile. Did anyone try this with prisma yet? It would be a kind of workaround at least.

larsivi:
[...]

PR #5028, opened 2020-03-01
Extend Docker build to allow for a folder called prerun_hook_files

larsivi:
If the folder exists in cwd, will be copied alongside the prerun_hook.sh script. Files inside will then be available to the script (from /app/prerun_hook_files).

References issue #4417

By putting certificates in this folder, I can use the Java keytool to add them to the store in the prerun_hook.sh script. These certificates are then successfully used when connecting to a Mongo database using ssl.


larsivi:
Closed this as it was against master, noticed that it is supposed to be against alpha.

larsivi:
Ok, not really sure about the alpha branch, all my testing has been against master, so reopening this pull request here.

larsivi:
Feel free to ignore this, as it probably makes more sense to do this (getting certs in) via building an entirely new Docker image based on the prisma one. I'm a bit new to Docker, so didn't see the obvious solution at first.

PR #5061, opened 2020-03-31
fix(README): fix all the links

janpio:
This PR tries to fix all the links in the README.

Still could use some updates to Prisma 2 wording though.


nikolasburk:
Thanks a lot Jan!

